PR #25, opened 2017-07-10
TFT_eSPI+parallel

konkrog:
Hi Bodmer,

This commit might not be suitable for the TFT_eSPI library...
    
What I've implemented is a parallel interface for ES32 to a TFT display with ILI_9163 driver.     I have not implemented the routines that are reading from the display... as I do nto need them I think.

I have tested all the example schetches on a ESP32 NodeMCU together with a TFT 1.8" screen from NewHavenDisplay (http://www.newhavendisplay.com/nhd18128160efcsxnf-p-9493.html)
    
The reason for selecting the display is that it was the only one with Sunligt Readable capability (1000cd), and it only has a parallel interface.
    
I would appreciate if the changes could somehow be incorporated into TFT_eSPI.

Please let me know what you think.
    
Regards,
Knut

### DIFF
--- a/TFT_eSPI.cpp
+++ b/TFT_eSPI.cpp
@@ -24,7 +24,9 @@
 //#include <limits.h>
 //#include "pins_arduino.h"
 //#include "wiring_private.h"
+#if !defined(ESP32_PARALLEL)
 #include <SPI.h>
+#endif
 
 // If it is a 16bit serial display we must transfer 16 bits every time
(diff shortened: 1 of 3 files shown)


Bodmer:
Hi,  you have done a good job with the changes.  I suspect that additional changes may be needed to ensure the original SPI bus code continues to work. I will have to check this before merging your code.

The library was originally intended for the ESP8266 which did not have enough pins to support a parallel display hence the name of the library contains SPI.

PR #31, opened 2017-08-01
ST7787 and 3-wire protocol support for TFT_eSPI

knielsen:
This pull request implements support for another display driver, the ST7787,
in your TFT_eSPI library.

With this request, I want to ask if you are interested in getting this
included in your upstream repository for others to use?

The implementation should be complete, all examples work. There is currently
only support for the ESP8266 though, ESP32 is not yet implemented.

There is some background on the project here, if you are interested:
http://kristiannielsen.livejournal.com/19836.html . Also, we have a bunch of
these displays lying around; if you are interested I can send you a couple
to play with^H^H^H^H test.

Datasheets for the display and the controller are here:

  http://www.glyn.co.nz/downloads/protected/EDT/Datasheets/TFT/ET024002DMU-RoHS%20ver%204%20-%202.4%20inch%20-%20240%20x%20320.pdf
  http://www.glyn.co.nz/downloads/protected/EDT/Datasheets/TFT/Controllers/ST7787_V1.1_20070614.pdf

The main difference with the ST7787 is that it only supports the 3-wire
serial interface - it has no DC (data/command) pin in serial mode. Instead
the D/C bit is sent as a 9th bit with every byte transmitted. Thus, all the
SPI-code needs to be modified to use 9 bits per byte. This means that the
patch adds a fair amount of code. On the other hand, the 3-wire interface is
also present in other controllers; this could be used to reduce pin count
(saving the need for the DC pin).

I tried to preserve the good performance of the library for the ST7787,
using the ESP8266 SPI hardware in a similar way. The need for 9-bit SPI
introduces some extra bit-stuffing overhead. On the other hand, since D/C is
inline in the SPI stream, multiple commands can be transmitted in one go by
the SPI hardware without manually flipping the DC pin, which might win back
a bit of performance (eg. setAddrWindow()).

The ESP32 is not supported at all for ST7787 for now. I could add it though,
if necessary (and of I can find an ESP32 to test with, shouldn't be hard).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,9 @@ $RECYCLE.BIN/
 # Windows shortcuts
 *.lnk
 
+# Backup files
+*~
+
 # =========================
 # Operating System Files
 # =========================
--- a/README.md
+++ b/README.md
@@ -1,10 +1,13 @@
 # TFT_eSPI
 
-An Arduino IDE compatible graphics and fonts library for ESP8266 and ESP32 processors with a driver for ILI9341, ILI9163, ST7735 and S6D02A1 based TFT displays that support SPI.
+An Arduino IDE compatible graphics and fonts library for ESP8266 and ESP32 processors with a driver for ILI9341, ILI9163, ST7735, ST7787 and S6D02A1 based TFT displays that support SPI.
 
 The library also supports TFT displays designed for the Raspberry Pi that are based on a ILI9486 driver chip with a 480 x 320 pixel screen. This display must be of the Waveshare design and use a 16 bit serial interface based on the 74HC04, 74HC4040 and 2 x 74HC4094 logic chips. A modification to these displays is possible (see mod image in Tools folder) to make many graphics functions much faster (e.g. 23ms to clear the screen, 1.2ms to draw a 72 pixel high numeral).
 
 The library supports SPI overlap so the TFT screen can share MOSI, MISO and SCLK pins with the program FLASH.
+
+The library supports the 3-wire mode of some display controllers (currently tested with ST7787 only, and currently only on ESP8266). This mode reduces pin count by not using the DC and MISO pins.
+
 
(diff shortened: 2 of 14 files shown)


Bodmer:
Thanks for this I will have a look at your code and see if it has any impacts on the existing implementation.  Unfortunately I do not have a ST7787 so this might make it tricky for me to support this display if bugs crop up.  I will think on this further.

Is there an online retailer that sells these displays?

knielsen:
I am not sure about the availability of the ST7787 display. We got a bunch of them as a donation. I think the cheap pre-made display breakout boards on ebay etc. are mainly using displays with the 4-wire protocol that is supported by existing libraries.

I will of course be happy to fix any bugs for the ST7787 that might turn up, but I understand the issue of you not having access to a display to test on. I can send you a couple of displays if you want to be able to test with them.

The main impact on existing code is added #ifdef sections for the new 3-wire protocol support. This code might also be useful to support the 3-wire protocol (to save a pin) on popular displays like ILI9341 and ST7735, which are very similar to the ST7787. This requires boards that are not hard-coded to use the 4-wire protocol (the IMx pins).

knielsen:
A couple references to these displays. Though the ST7787 does seem less commonly seen that the other similar controllers:

  http://forum.arduino.cc/index.php?topic=26287.0

  http://www.ebay.es/itm/161655184635

Bodmer:
I have not found a retailer that sells the ST7787 in the SPI flavour, the only ones I have seen are relatively rare Arduino shields with an 8 bit parallel interface, so I am putting the merge on hold for now.

knielsen:
Yeah, the idea with merging this upstream was that others could benefit. If no-one else is using these kinds of displays, merging probably makes less sense.

I can keep the fork with ST7787/3-wire support on my github, so that if some user turns up they will be able to find it and use it.

Thanks,

 - Kristian.

Bodmer:
Thanks for your input, if these displays become commonly available then I will introduce the support for the chip. In the meantime I will close this.

PR #36, opened 2017-09-10
yet another ST7735 variant

MarkyAD:
Found another  ST7735, got these at MiniArduinoStore on ebay. Blue PCB with single 8-pin header. It inits like other green tab but doesn't need the CGRAM_OFFSET. None of the other configs (red/black/green/etc) worked until I added this, they either had artifacts or gave just a blank screen.


MarkyAD:
Found out why this screen didnt work with an already defined config. There is a small typo in ST7735_Defines.h, so its actually not a new variant, it works with red tab and this patch is not needed.

PR #37, opened 2017-09-11
fix for ST7735_REDTAB

MarkyAD:
Sorry for my confusion over the last commit. This one-liner is all that was needed to get these displays working.


Bodmer:
Thanks for finding this.

PR #38, opened 2017-09-16
Conversion from 24 bit and css color notation

emilekm2142:
I have opened an issue a few days ago about the colors from hex value and struggled a bit to get a working solution. Therefore I would like to add this feature so other beginners could save some time.
Now you can get a 16 bit color that pretty much reassembles the 24 bit one with either
`tft.color565From888(0xffffff);` or `tft.color565FromWebHex("#ffffff");`

### DIFF
--- a/TFT_eSPI.cpp
+++ b/TFT_eSPI.cpp
@@ -2710,7 +2710,27 @@ uint16_t TFT_eSPI::color565(uint8_t r, uint8_t g, uint8_t b)
 {
   return ((r & 0xF8) << 8) | ((g & 0xFC) << 3) | (b >> 3);
 }
+uint16_t TFT_eSPI::color565FromWebHex(const char* color){
+		if (color[0] == '#') { 
+			//remove first char
+			String c = String(color);
+			c = c.substring(1);
+			char l[9];
+			for (int i = 0; i<9; i++)
+				l[i] = '\0';
+			c.toCharArray(l, 9);
+			color = l;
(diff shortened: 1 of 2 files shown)


Bodmer:
If you have a 24 bit RGB colour 0xAABBCC then the existing color565 function could be used just be splitting out the 3 colour bytes, eg:

color16 = tft.color565(0xAA, 0xBB, 0xCC);

I will have a look at your code, I prefer to use function overloading so that that a single named function handles different formats. eg support:

tft.color565(0xffffff);
and
tft.color565("#ffffff");

This avoid the situation I see in some libraries where there are lots of functions with subtle differences in the names that users then find difficult to remember :-)

Bodmer:
I had a look at your code and see a couple of issues:

A 24 bit colour is being passed to the **color565From888(0xffffff)**; function, but the [code here](https://github.com/Bodmer/TFT_eSPI/pull/38/commits/24167098eea345b92d0d939b54ccfbc1a7a40dda#diff-a30868faa3e646e6ae51eab6ef6450f3R2728) only expects a 16 bit value.

Ideally error checking should be incorporated into the **color565FromWebHex("#ffffff");** function to handle the case where the user does not include a 6 character colour code. Otherwise the code would read off the end of the array and generate spurious outputs that would confuse the user. The difficulty of handling situation like this is why the hex passing method is better than passing character arrays as that will always correctly use the value passes.

PR #45, opened 2017-10-11
ESP32 Change use of SPI transactions

beegee-tokyo:
Using the #define SUPPORT_TRANSACTIONS to initialize SPI on ESP32 with or without transaction support.

Digging a little bit more into my problem with ArduinoOTA and TFT_eSPI. As transactions eat up a little performance I tried to use SPI on the ESP32 without transactions. Instead of 
_**`SPI.beginTransaction(SPISettings(SPI_FREQUENCY, MSBFIRST, SPI_MODE0));`**_
I initialized SPI with
```
SPI.begin(TFT_SCLK,TFT_MISO,TFT_MOSI);
SPI.setFrequency(SPI_FREQUENCY);
SPI.setBitOrder(MSBFIRST);
SPI.setDataMode(SPI_MODE0);
```
This initializations works as well and does not use SPI transactions.  And transactions are only needed if several devices are connected to the same bus.

A small code change gives the user the same possibilities than on an ESP8266, to choose if SPI transactions are needed and used.
```
  #ifdef ESP32 // Unlock the SPI hal mutex and set the lock management flags
    #ifdef SUPPORT_TRANSACTIONS
      SPI.beginTransaction(SPISettings(SPI_FREQUENCY, MSBFIRST, SPI_MODE0));
      inTransaction = true; // Flag to stop intermediate spi_end calls
      locked = false;       // Flag to stop repeat beginTransaction calls
    #else
      SPI.begin(TFT_SCLK,TFT_MISO,TFT_MOSI);
      SPI.setFrequency(SPI_FREQUENCY);
      SPI.setBitOrder(MSBFIRST);
      SPI.setDataMode(SPI_MODE0);
    #endif
  #endif
```
Tested on my ESP32 Dev module with the ILI9163 TFT display.

### DIFF
--- a/TFT_eSPI.cpp
+++ b/TFT_eSPI.cpp
@@ -192,9 +192,16 @@ void TFT_eSPI::init(void)
   SPI.setFrequency(SPI_FREQUENCY);
 
   #ifdef ESP32 // Unlock the SPI hal mutex and set the lock management flags
-    SPI.beginTransaction(SPISettings(SPI_FREQUENCY, MSBFIRST, SPI_MODE0));
-    inTransaction = true; // Flag to stop intermediate spi_end calls
-    locked = false;       // Flag to stop repeat beginTransaction calls
+    #ifdef SUPPORT_TRANSACTIONS
+      SPI.beginTransaction(SPISettings(SPI_FREQUENCY, MSBFIRST, SPI_MODE0));
+      inTransaction = true; // Flag to stop intermediate spi_end calls
+      locked = false;       // Flag to stop repeat beginTransaction calls
+    #else
+      SPI.begin(TFT_SCLK,TFT_MISO,TFT_MOSI);
+      SPI.setFrequency(SPI_FREQUENCY);
+      SPI.setBitOrder(MSBFIRST);
+      SPI.setDataMode(SPI_MODE0);
+    #endif
   #endif
 
 #endif


Bodmer:
Thanks for investigating this, I will update the library with your suggested change
Can you cobfirm OTA works OK with this modification.

beegee-tokyo:
Yes I can confirm that there is no more problem with OTA.
I tested with **`#define SUPPORT_TRANSACTIONS`** enabled and disabled.

MarkyAD:
I tried this patch (ESP32+ILI9341) but I got the same results as @beegee-tokyo did originally... screen garbage and OTA failures/guru meditation errors. Not sure why, everything works nicely without OTA.

beegee-tokyo:
@MarkyAD the original code for ESP32 starts SPI always with using transactions, when I found the problem the first time, I could solve it with putting `SPI.endTransaction()` into `ArduinoOTA.onStart([]() {...}`, but then I couldn't use the display anymore during OTA. I wanted to use the display during OTA to show the progress and evtl. errors if they occur.

That's why I tried to use the lib without transactions at all.

Are you using Arduino IDE or PlatformIO? 
Are you using the master framework (branch origin/master V0.10.0) or the staging version of the framework (current version aba1d7d)? OTA didn't work with the master framework at all. I had to switch to the staging version to get it to work. [Using Arduino Framework with Staging version](http://docs.platformio.org/en/latest/platforms/espressif32.html#using-arduino-framework-with-staging-version)

MarkyAD:
Ok maybe thats it, Im using the current origin/master on Arduino IDE. I too am using tft_eSPI function calls from within the progress event but I commented that out and it still does it.

beegee-tokyo:
@MarkyAD not using the display still crashed my ESP32 as well. Only the change in the TFT_eSPI lib and the switch to staging version solved the problem. Here a small [YouTube video](https://youtu.be/zlRhqLJoqUM) showing it at work.

MarkyAD:
Nice, I have it working in a similar fashion on esp8266+ILI9341. So only the platformIO staging version works on esp32 currently.

beegee-tokyo:
Yes, see this [issue](https://github.com/espressif/arduino-esp32/issues/612) on arduino-esp32.

Bodmer:
As far as I can see this suggested change will not make any difference as it is within the #ifndef... #endif at line [194 to 206.](https://github.com/Bodmer/TFT_eSPI/blob/master/TFT_eSPI.cpp#L194-L206). Since the code is only compiled if SUPPORT_TRANSACTIONS has been defined!

Bodmer:
OK, I see that the code change does have an effect. It looks like OTA updates don't like the mutex being permanantly grabbed by the TFT library. Grabbing the mutex permanently improves performance on the ESP32 but this is at the expense of failing if OTA is used. This means that it is best to always enable transacrion support on the ESP32. In practice the performance reduction is small so I will make this change in the next few days.

Bodmer:
I have changed the code to always request SPI port access via the hal mutex. This should now be compatible with OTA.

Thanks for reporting the issue.

PR #57, opened 2017-11-26
Fix width/height override issue when using setRotation

turiphro:
This commit fixes an issue when setting screen width
and height from the constructor (e.g., user sketch),
followed by setRotation(i). Previously, setRotation
used TFT_WIDTH and TFT_HEIGHT, ignoring user overrides.

### DIFF
--- a/TFT_Drivers/ILI9163_Rotation.h
+++ b/TFT_Drivers/ILI9163_Rotation.h
@@ -7,8 +7,8 @@
   switch (rotation) {
     case 0:
       writedata(TFT_MAD_MX | TFT_MAD_MY | TFT_MAD_BGR);
-      _width  = TFT_WIDTH;
-      _height = TFT_HEIGHT;
+      _width  = _width_orig;
+      _height = _height_orig;
 	  #ifdef CGRAM_OFFSET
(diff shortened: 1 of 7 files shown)


Bodmer:
Thanks for pointing this issue out.

The setRotation() will need to be moved out of the constructor, as at that point the SPI bus has not been configured.  It could be called towards the end if the init() function.

Another potential issue is that the xxxx_init.h files configure the display width, height and rotation too in the control registers of the TFT driver, thus there could be a conflict between the user settings and the init settings.

I think calling calling set rotation is safe towards the end of init() would be OK, but I will need to consider this further before merging.

turiphro:
The original values `_weight` and `_height` can also be set explicitly, like before; it's just that calling setRotation seemed cleaner. Changed to explicit assignment to prevent unforeseen side-effects, see updated commit.

Can you elaborate on where the _init.h files configure the display w/h/r? None of these files currently seem to be using the `TFT_WIDTH` and `TFT_HEIGHT` variables, only hard-coded variables.

Bodmer:
The init values setup a window and will be later over-written so are not important.

I have merged. Thanks for your input.

Issue #62, opened 2017-12-06
[Enhancement] Shift a sprite left or right


beegee-tokyo:
I am working on a function to display temperature values as a graph on the display. And I wanted to shift(scroll???) the graph when the screen is filled and add the new data at the end. 
My display (ILI9163) has no function to shift the screen content or read back the content of the screen.

I decided to work with a sprite, but the library has neither a function to read back the sprite content for changing nor a shift(scroll???) function.

To achieve what I want, I added 2 sprite functions to the library, **`shiftRectLeft()`** and **`shiftRectRight()`**
Maybe you want to add these to the library.
Tested only with an [ILI9163 128x128 TFT](https://www.elecrow.com/144-128x-128-tft-lcd-with-spi-interface-p-855.html)  display and an ESP32.

The declarations:
```cpp
           // Shift the sprite <shift> pixels to the left and fill right side with <color>
  void     shiftRectLeft(uint32_t x, uint32_t y, uint32_t w, uint32_t h, uint32_t shift, uint32_t color);
           // Shift the sprite <shift> pixels to the right and fill left side with <color>
  void     shiftRectRight(uint32_t x, uint32_t y, uint32_t w, uint32_t h, uint32_t shift, uint32_t color);
```
The functions:
```cpp
/***************************************************************************************
** Function name:           shiftRectLeft
** Description:             shifts sprite left and fill right side with given color
***************************************************************************************/
void  TFT_eSprite::shiftRectLeft(uint32_t x, uint32_t y, uint32_t w, uint32_t h, uint32_t shift, uint32_t color)
{
  if ((x > _iwidth) || (y > _iheight) || (w == 0) || (h == 0) || !_created) return;

  for (uint32_t yp = y; yp < y + h; yp++)
  {
    for (uint32_t xp = x + shift; xp < x + w; xp++)
    {
      _img[xp-shift + yp * _iwidth] = _img[xp + yp * _iwidth];
    }
    for (uint32_t xp = x + w; xp < x + w + shift; xp++)
    {
      if (_bpp16)
      {
          _img[xp-shift + yp * _iwidth] = (uint16_t) (color >> 8) | (color << 8);
      }
      else
      {
        _img[xp-shift + yp * _iwidth] = (uint8_t)((color & 0xE000)>>8 | (color & 0x0700)>>6 | (color & 0x0018)>>3);
      }
    }
  }
}

/***************************************************************************************
** Function name:           shiftRectRight
[...]

PR #63, opened 2017-12-06
[Enhancement] Shift a sprite left or right

beegee-tokyo:
Pull request for enhancement as posted in #62 

In addition this includes a function to read back the content of a sprite for processing into a buffer.
**`void  TFT_eSprite::readRect(uint32_t x, uint32_t y, uint32_t w, uint32_t h, uint16_t *data)`**


Bodmer:
See response to issue #62 

Also I am adding a function to return a pointer to the sprite memory block when the sprite is created, this means you will then have direct read/write access to the stored sprite memory area and do not need to make a copy with readRect(). If you did need to make a copy then memcpy() can be used as this is super fast.

PR #67, opened 2017-12-16
Allow externally defined configuration rather than User_Setup.h modifications in library

atanisoft:
Adding ability to load settings from the calling program rather than modifying the library for each project that uses it. This is mostly for PlatformIO IDE which manages the libraries differently than Adruino IDE.

### DIFF
--- a/User_Setup_Select.h
+++ b/User_Setup_Select.h
@@ -15,7 +15,7 @@
 // Customised User_Setup files are stored in the "User_Setups" folder.
 
 // Only ONE line should be uncommented.  Add extra lines and files as needed.
-
+#ifndef USER_SETUP_LOADED
 #include <User_Setup.h>           // Default setup is root library folder
 
 //#include <User_Setups/Setup1_ILI9341.h>  // Setup file configured for my ILI9341
@@ -31,8 +31,7 @@
 //#include <User_Setups/Setup11_RPi_touch_ILI9486.h>  // Setup file configured for my stock RPi TFT with touch
 
 //#include <User_Setups/SetupX_Template.h> // Setup file template for copying/editting
-
-
+#endif
 
 /////////////////////////////////////////////////////////////////////////////////////
 //                                                                                 //


Bodmer:
Great! I have just loaded PlatformIO to try and I can see why this change is neat, simple and a useful update.

Thanks.

atanisoft:
Thanks! I was running in circles for a bit trying to get a display working on an esp32 and it wouldn't work. I modified the library to try a few options before arriving at this one.

Bodmer:
I will make some changes to make it clear which user setups are for the ESP32.

beegee-tokyo:
Didn't work for me.
I changed **`Update User_Setup_Select.h`** as shown in the pull request.
Then I tried to define the display settings in my include file just before I include **`TFT_eSPI.h`**:
```cpp
...
#define USER_SETUP_LOADED 1
#define ILI9163_DRIVER
#define TFT_WIDTH  128
#define TFT_HEIGHT 160
#define TFT_MISO 19
#define TFT_MOSI 23
#define TFT_SCLK 18
#define TFT_CS    5  // Chip select control pin
#define TFT_DC    19  // Data Command control pin
#define TFT_RST  -1  // Set TFT_RST to -1 if display RESET is connected to ESP32 board RST
#define LOAD_GLCD   // Font 1. Original Adafruit 8 pixel font needs ~1820 bytes in FLASH
#define SPI_FREQUENCY  27000000 // Actually sets it to 26.67MHz = 80/3
#include "TFT_eSPI.h"
...
```
But the user settings are not used. What am I doing wrong?
PlatformIO version: Home 0.5.3·Core 3.5.0rc10
ESP32 platform: Espressif 32 (Stage) version c9fe5b9
Atom version 1.23.1 x64

atanisoft:
Add them to your platformio.ini file via build_flags as -DTFT_RST=-1 etc.

beegee-tokyo:
Thank you! Now it works.

PR #76, opened 2018-01-14
Add second variant of 128x128 CGRAM offset

vanklompf:
Default setting is still OFFSET_FROM_END.
Also if CGRAM_OFFSET is defined without value, it should still work as before.
Now works on my LCD.


Bodmer:
Thanks for this. A few comments:
CGRAM_OFFSET is used in ILI_9163_Init.h to set the intital offset, this will get fixed though when setRoration() is called.

The ILI9163 controller can drive 162x132 displays so I suspect there will be other 2 pixel offset combinations that may need to be supported as is experienced with the ST7735 controllers. As illustrated by this snippet of code:
`  switch (rotation) {
    case 0:
     if (tabcolor == INITR_BLACKTAB) {
       writedata(TFT_MAD_MX | TFT_MAD_MY | TFT_MAD_RGB);
     } else if(tabcolor == INITR_GREENTAB2) {
       writedata(TFT_MAD_MX | TFT_MAD_MY | TFT_MAD_RGB);
       colstart = 2;
       rowstart = 1;
     } else if(tabcolor == INITR_GREENTAB3) {
       writedata(TFT_MAD_MX | TFT_MAD_MY | TFT_MAD_BGR);
       colstart = 2;
       rowstart = 3;
     } else if(tabcolor == INITR_GREENTAB128) {
       writedata(TFT_MAD_MX | TFT_MAD_MY | TFT_MAD_MH | TFT_MAD_BGR);
       colstart = 0;
       rowstart = 32;
`

So I am considering a change that will allow any combination of offsets to be accommodated for both the ST7735 controllers and the ILI9163, e.g for your display.

R0_OFFSET_X 0
R0_OFFSET_Y 32
R1_OFFSET_X 32
R1_OFFSET_Y 0
R2_OFFSET_X 0
R2_OFFSET_Y 0
R3_OFFSET_X 0
R3_OFFSET_Y 0

vanklompf:
Ok, makes sense.
I will make updates to PR.

Bodmer:
This looks goid. I will resolve conflcts and incorporate the changes.

janoist1:
Any news with this?

Bodmer:
The conflicts are messy to resolve and I am not keen on a common rotation function so this needs reworking and some further thought.  As Summer approaches I find less time to to tweaks so this one will have to wait a while.

vanklompf:
@janoist1 
If you are using PlatformIO you can try my fork:
https://github.com/vanklompf/TFT_eSPI
I have reworked configuration so that every variant can be injected from platformio.ini config without need to edit library source files. 
Example is here: https://github.com/vanklompf/TFT_Demo
It should also work for non-PlatformIO builds, but I'm using it only with PlatformIO.

Bodmer:
Time has moved on and so this change will not be made in this format.

PR #85, opened 2018-01-27
Added public resetLcd() function

MarcFinns:
Allows to reset the panel without messing with the SPI interface. Useful in cases where the panel needs a reset and calling the init() function would create issues as it resets the GPIOs related to SPI. For example, if MISO is not used and the related GPIO is driving something else


Bodmer:
I've implemented and equivalent approach.

Thanks.

PR #87, opened 2018-02-01
PartialMode and Idle mode support for ILI9163

vanklompf:
Support of partial mode, where only part of screen is active.  Can be use to reduce power consumption.
Based on ILI9163 datasheet:
https://www.rockbox.org/wiki/pub/Main/SonyNWZE370/ILI9163.pdf


Bodmer:
Tests indicate the power savings are tiny compared to the power consumed by the backlight so I do not see any real benefit of adding this.  Thanks for yor suggestion.^

PR #88, opened 2018-02-03
Added support for M5Stack

tonywestonuk:
Added support for M5Stack - http://www.m5stack.com/

Involved changing the rotations, screen resolution (320x240 rather than 240x320), and adding a LCD LED control

### DIFF
--- a/TFT_Drivers/M5Stack_Defines.h
+++ b/TFT_Drivers/M5Stack_Defines.h
@@ -1,0 +1,126 @@
+// Change the width and height if required (defined in portrait mode)
+// or use the constructor to over-ride defaults
+#define TFT_WIDTH  320
+#define TFT_HEIGHT 240
+
+
+// Color definitions for backwards compatibility with old sketches
+// use colour definitions like TFT_BLACK to make sketches more portable
+#define ILI9341_BLACK       0x0000      /*   0,   0,   0 */
+#define ILI9341_NAVY        0x000F      /*   0,   0, 128 */
+#define ILI9341_DARKGREEN   0x03E0      /*   0, 128,   0 */
+#define ILI9341_DARKCYAN    0x03EF      /*   0, 128, 128 */
+#define ILI9341_MAROON      0x7800      /* 128,   0,   0 */
+#define ILI9341_PURPLE      0x780F      /* 128,   0, 128 */
+#define ILI9341_OLIVE       0x7BE0      /* 128, 128,   0 */
+#define ILI9341_LIGHTGREY   0xC618      /* 192, 192, 192 */
+#define ILI9341_DARKGREY    0x7BEF      /* 128, 128, 128 */
+#define ILI9341_BLUE        0x001F      /*   0,   0, 255 */
+#define ILI9341_GREEN       0x07E0      /*   0, 255,   0 */
+#define ILI9341_CYAN        0x07FF      /*   0, 255, 255 */
+#define ILI9341_RED         0xF800      /* 255,   0,   0 */
+#define ILI9341_MAGENTA     0xF81F      /* 255,   0, 255 */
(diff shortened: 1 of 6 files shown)


tonywestonuk:
This was a copy/paste/modify from the ili9341 driver. It works ok, though maybe it would be nicer in future to be able to carry on using the stock ILI9341 driver, but be able to include additional configs for different types of LCD that uses this?  Otherwise will end up with lots of duplication.

Bodmer:
You can include tft.setRotation(1) or tft.setRotation(1) in the sketch setup to swap to landscape orientation so it is not clear to me why new driver files would be needed.

tonywestonuk:
Its easier if I show you - Screen shots attached with stock driver - The first 3 are different rotations, 0 , 1 ,2 - The rotation rotates the image in the 240x320 resolution, not the viewport itself.  Changing the resolution to 320x240 in Defines.h fixes this.

The final image is what the stock rotations.h does - everything is reversed! - like a mirror - regardless of what rotation is set.
img_5429
img_5426
img_5427
img_5428

tonywestonuk:
img_5431

reaper7:
yes, it's true,  everything is mirrored on all available rotations.
what about new(additional) rotations for m5stack?

TFT_eSPI\TFT_Drivers\ILI9341_Rotation.h
line 4 from:
```
rotation = m % 8; // Limit the range of values to 0-7
```
to:
```
rotation = m % 12; // Limit the range of values to 0-7
```
and new cases:
```
  // These next rotations are for m5stack
    case 8:
      writedata(TFT_MAD_MX | TFT_MAD_MY | TFT_MAD_BGR);
      _width  = _height_orig;
      _height = _width_orig;
      break;
    case 9:
      writedata(TFT_MAD_MV | TFT_MAD_MX | TFT_MAD_BGR);
      _width  = _width_orig;
      _height = _height_orig;
      break;
    case 10:
      writedata(TFT_MAD_BGR);
      _width  = _height_orig;
      _height = _width_orig;
      break;
    case 11:
      writedata(TFT_MAD_MY | TFT_MAD_MV | TFT_MAD_BGR);
      _width  = _width_orig;
      _height = _height_orig;
      break;
```

Bodmer:
Thanks I will add support for the M5Stack. I will consider further how to do this whilst making sure existing sketches run without modification.

reaper7:
so...second option (without additional cases)
but currently macro name for M5Stack is not entirely compatible with arduino standard:
https://github.com/m5stack/M5Stack/issues/41
```
#ifdef ARDUINO_M5Stack-Core-ESP32 //orientations 0-3 for M5Stack (as above)
case 0:
.
.
.
#else                                                    //existing orientations 0-3 (from current code)
case 0:
.
.
.
#endif
```

Bodmer:
I have added support for the M5Stack but cannot test it. I have ordered an M5Stack to play with!

Rotation 0 is portrait mode so you will need to call setRotation(1) in your sketch to orient the display.

Thanks for your input to this change.

reaper7:
I tested almost all examples, for me rotations 0-3 works very well!

Two comments:
1. about rotations 4-7 (I do not use them but I checked, please see if this is how it should be):
- 4 looks like 2
- 5 looks like 0 but mirrored and probably swapped W with H
- 6 looks like 2 but mirrored
- 7 looks like 3 but mirrored

2. keypad example (yes, I know that m5stack does not have touch ;) ) -> compilation error: SPIFFS not declared 
it seems that esp32 needs in this example additional #include "SPIFFS.h"

tonywestonuk:
Closing pull request as no longer needed.  Thank you  Bodmer / Reaper7 :-)

Bodmer:
Mirrored rotations 4-7 were wrong so I have corrected these.

PR #91, opened 2018-02-08
Move TFT_SPI_OVERLAP define near pins, Add GPIOs.

JamesNewton:
Very minor change, but it would have saved me hours of work so maybe it will help some other lug. 

Move the TFT_SPI_OVERLAP define up near the PINS section so it is more obvious.
Add GPIO pins to documentation.


Bodmer:
Addressed in latest setp files.

PR #92, opened 2018-02-08
Initial commit

JamesNewton:
Setup for ILI9341 in overlap mode (sharing ESP-8266 SPI0 port)


Bodmer:
Addressed in latest setup example files.

PR #93, opened 2018-02-08
Add Setup13_ILI9341_Overlap

JamesNewton:
The file I actually suggested was named Setup12_ILI9341_Overlap.h so that will need to get renamed.


Bodmer:
I have addresses this is the latest update by making all setup files consistent and based on the same template.  Thanks for your input.

PR #122, opened 2018-04-11
Update Sprite.cpp

rdts:
Hi,
I found a bug.

_width and _height need to be initialized properly.

That is because they are used in drawString(). Lines 4118 and 4120 in TFT_eSPI.cpp.


Bodmer:
Thanks!

Bodmer:
I had problems with the suggested fix. I have fixed it now though.
Thanks for taking the time to report this bug.

PR #164, opened 2018-06-24
 added fontHeight() function with no parameter

maxpautsch:
added the fontHeight() function with no parameter to be consistent with the textWith() function and others. no parameter means default font.


Bodmer:
Thanks, I have included this in the latest update.

PR #182, opened 2018-08-01
Added ST7789 initialization

amakukha:
I added ST7789 initialization sequence according to ST7789 datasheet. 
Tested with display JLX240-00302 (320x240, 2.4") which uses ST7789V display driver.

It looks like setting SPI_MODE3 for ST7789 is not needed, so I deleted that part.

### DIFF
--- a/TFT_Drivers/ST7789_Defines.h
+++ b/TFT_Drivers/ST7789_Defines.h
@@ -1,7 +1,11 @@
 // Change the width and height if required (defined in portrait mode)
 // or use the constructor to over-ride defaults
-#define TFT_WIDTH  240
-#define TFT_HEIGHT 240
+#ifndef TFT_WIDTH
+  #define TFT_WIDTH  240
+#endif
+#ifndef TFT_HEIGHT
+  #define TFT_HEIGHT 320
+#endif
 
 #define CGRAM_OFFSET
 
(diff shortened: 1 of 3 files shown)


Bodmer:
Thanks for you input.

I have an ST7789 board that has the CS line tied low permanently, this board will not work with SPI_MODE0 but does with SPI_MODE3.

Can you test if the same happens with your display if you tie the CS line permanently low? If you can, then let me know your findings.

Also does your board work OK with the CS line driven and with SPI_MODE3?

amakukha:
I don't think CS line is related to SPI mode. CS low simply means that the slave device is permanently selected. This makes sense in our design (we are building [“ESP32 WiPhone”](https://hackaday.io/project/159811-esp32-wiphone)) because display uses a dedicate SPI line, so there is always only one slave - the screen.

Also, according to the ST7789V datasheet:

> At the falling edge of CSX, SCL can be high or low. SDA is sampled at the rising edge of SCL.

I think, this means that both SPI_MODE0 and SPI_MODE3 should both work. 

In my testing, SPI_MODE0, SPI_MODE3 and SPI_MODE2 all work no matter how CS is connected (tied low or driven). The frame rate appears to be the same in these three modes. The only mode that doesn't work for me is SPI_MODE1 (no matter how CS is connected).

So by some luck SPI_MODE2 also works for me. 

Certainly, it is strange that SPI_MODE0 doesn't work for you. I guess, it might be caused by impedance differences in your specific setup. So, in general, it might be wise to allow users to select SPI mode.

amakukha:
Oh, I actually remembered one dev board I have that had problems with the ST7789 screen: it required a soft reset in order for the screen to work properly. Otherwise, the screen would always stay white.

But this bug disappeared after I changed SPI mode to SPI_MODE3 :sun_with_face: So finally, I think it is safer to leave the default SPI_MODE3 for ST7789 driver.

PR #211, opened 2018-09-27
Fix for smooth font anti aliasing in sprites

pteal:
Currently, smooth fonts are antialiased differently depending on whether they are drawn directly on a TFT or drawn on a sprite and then pushed to the TFT.  In my opinion, the anti aliasing should be consistent, regardless of where the text is being drawn. 

Digging into the libraries, I noticed that drawGlyph() in Sprite.cpp overrides drawGlyph() in Smooth_font.cpp.  For some reason, in Sprite.cpp, drawPixel() was only being called if the alpha value of the pixel was greater than 127.  This produced weaker anti aliasing.  Removing the "if" statement and calling drawPixel() regardless of the alpha value of that pixel, resolves the inconsistency and produces great looking text!

old-bad
The image above shows the anti aliasing differences.  

The images below show white text written on a white background, but with black used for the antialiasing background color to highlight the antialiased pixels.   

old
The image above shows text drawn directly to the TFT on the top line and text drawn on a sprite and then pushed to the TFT on the second line.

new
The image above shows the results after making the proposed change to Sprite.cpp.

### DIFF
--- a/Extensions/Sprite.cpp
+++ b/Extensions/Sprite.cpp
@@ -1525,7 +1525,7 @@ void TFT_eSprite::drawGlyph(uint16_t code)
           if (pixel != 0xFF)
           {
             if (dl) { drawFastHLine( xs, y + this->cursor_y + this->gFont.maxAscent - this->gdY[gNum], dl, fg); dl = 0; }
-            if (pixel>127) drawPixel(x + this->cursor_x + this->gdX[gNum], y + this->cursor_y + this->gFont.maxAscent - this->gdY[gNum], alphaBlend(pixel, fg, bg));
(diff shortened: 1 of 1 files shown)


Bodmer:
Thanks for spotting this!

The >127 threshold was supposed to be a temporary hack for Sprites that are used for ePaper displays. It allows the antialiased fonts to render reasonably for 1 bpp Sprites, so a test needs to be added. I forgot to replace the hack with the correct code.

I will update the library soon.

pteal:
I knew there must have been a good explanation for that bit of code.  Thanks for all your hard work on this library!

Bodmer:
The library has been updated to correct this.

There are a few possible methods of rendering the smooth fonts so I have created some [new smooth font examples](https://github.com/Bodmer/TFT_eSPI/tree/master/examples/Smooth%20Fonts) to show how it can be done.

PR #243, opened 2018-11-15
Example and short description for TFTShape extension

androdlang:
I think it's a good idea to start with the docs/examples...


androdlang:
Btw. thank you for the screenshot client, this is very helpful for the doku :)

Bodmer:
The pull is rather cluttered with historical changes and looks incomplete.

I suggest you develop your branch of the library, I will then test and comment on the branch until it has progressed to the point of being ready for release.

The aim is to minimise the number of incremental changes to the master copy of TFT_eSPI.

Bodmer:
I have had some more thoughts on this.  Up to this point most of the pull requests for TFT_eSPI have been relatively minor or bug fixes and thus I can manage the consequences of pulling the changes in.

In this tft2D case the additions are more significant and thus I do not want to end up having to maintain the extensions, documentation or handle any issues raised.

The way around this is for the extensions you propose to be in a completely different library which you maintain and update as you wish. That way the library development is not so heavily linked, you are free to make additions and changes and potentially your library could be used with other display libraries. I will include one or two examples that require the user to load your library, you are then free to add further examples to your library.

I suggest that the way forward is therefore to develop the tft2D extensions as a new Arduino library that interfaces with TFT_eSPI. There is a least one GUI library that works this way:

https://github.com/ImpulseAdventure/GUIslice

The way to handle the interface is for the library init code to accept a pointer to the tft class, in the same way as the Sprite class. You can use that pointer to call the functions in the TFT_eSPI library. I will let you choose the name, but maybe TFT_2DFX or TFT_eFX?

PR #321, opened 2019-02-27
Added D-Duino32-XS 'DSTIKE' ST7789 240x240

tobozo:


Bodmer:
I have added a second basic init option for the ST7789, this is the original code that was built into the library and was replaced by a more extensive init sequence. It appears to be identical to your suggested change.

PR #327, opened 2019-03-13
Add setFS function. ( Support for file systems other than SPIFFS. )

lovyan03:
SmoothFont can be read from SD in the environment where SD is available like M5Stack.
(default : SPIFFS)
```
#include <SD.h>
TFT_eSPI tft = TFT_eSPI();
void setup() {
  tft.init();
  tft.setFS(SD);
  tft.loadFont("fileName");
...
```

### DIFF
--- a/Extensions/Smooth_font.cpp
+++ b/Extensions/Smooth_font.cpp
@@ -77,12 +77,12 @@ void TFT_eSPI::loadFont(String fontName)
    unloadFont();
     
   // Avoid a crash on the ESP32 if the file does not exist
-  if (SPIFFS.exists("/" + fontName + ".vlw") == false) {
+  if ((_pFS ? *_pFS : SPIFFS).exists("/" + fontName + ".vlw") == false) {
     Serial.println("Font file " + fontName + " not found!");
     return;
   }
 
-  fontFile = SPIFFS.open( "/" + fontName + ".vlw", "r");
+  fontFile = (_pFS ? *_pFS : SPIFFS).open( "/" + fontName + ".vlw", "r");
 
   if(!fontFile) return;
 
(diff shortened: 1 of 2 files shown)


Bodmer:
Thanks for this. I will think further on how to add this capability. I suspect rendering the fonts off SD card will be slow, I will run some tests. One option on the To Do list is to buffer the loaded font in PSRAM.

lovyan03:
Thank you.
Font files tend to be huge in Japanese and Chinese. Please be cautious.

lovyan03:
I tested the drawing speed with SPIFFS and SD.
https://twitter.com/lovyan03/status/1117686353393635328
vlw file size is 167 kBytes.
SD was four times faster than SPIFFS.

I would like to use smooth_font on SD.
best regard.

Bodmer:
I have added SD card support for the ESP32, the font metrics also use PSRAM if it is available and enabled.

A new smooth font example has been added.

I found an SDHC card works but is ~3x slower than SPIFFS, so a microSD is recommended.

Let me know if you have any issues.

Bodmer:
Can you post your test program so I can see why you get such good performance?

lovyan03:
I used this font to try Japanese display.

https://github.com/watako-lab/M5Stack_bin/tree/master/M5Stack_SmoothFont/sd_card

ud36.vlw

With this font, SPIFFS was extremely slow.

Bodmer:
OK, I understand, the issue is that the file seek takes a long time with SPIFFS [as noted here](https://github.com/Bodmer/TFT_eSPI/blob/master/Extensions/Smooth_font.cpp#L473).

For small font files (<28kbytes) SPIFFS is quite fast because it appears to buffer the entire font file in RAM.

Thanks.

Issue #365, opened 2019-05-21
Touch support for dynamic setRotation()


ImpulseAdventure:
Thanks again for the great work on TFT_eSPI!

In testing the TFT_eSPI **integrated XPT2046 touch handler**, I noticed that the touch remapping (per calibration) does not appear to be updated following runtime rotation changes.

The default calibration (**Touch_calibrate**) captures the parameter set from rotation=1. The Touch_calibrate sketch demonstrates correct touch mapping to the display in this same orientation (rotation=1).

### Observation
However, if `getTouch()` is called after updating the display orientation via `setRotation()`, the touch remapping no longer appears to be rotated correctly:
- **rotation=0**: Touch is rotated -90deg clockwise
- **rotation=1**: Matched
- **rotation=2**: Touch is rotated +90deg clockwise
- **rotation=3**: Touch is rotated +180deg clockwise

As you are aware, other displays may exhibit a different remapping transformation than reflected above.

### Proposed Solution
My assumption from looking at the code is that while TFT_eSPI supports dynamic rotation for the display, dynamic rotation is not currently supported for touch. The touch rotation transform is only updated in the course of establishing the calibration parameters (`setTouch()`, `calibrateTouch()`).

Perhaps it may be possible to define a `touchRotate()` function that recomputes the `touchCalibration_rotate` / `touchCalibration_invert_x` / `touchCalibration_invert_y` parameters based on deltas in `rotation`. Then, upon calls to the `tft.setRotation(rotation)`, a call to the `touchRotate()` is also made.

### Testcase
I have attached a modified **Touch_calibrate-rotate** sketch to help you recreate the issue. After 200 points are collected, the display orientation is advanced.
[Touch_calibrate-rotate.zip](https://github.com/Bodmer/TFT_eSPI/files/3202863/Touch_calibrate-rotate.zip)

### Background details
- **MCU**: SparkFun ESP32 Thing
- **TFT**: 2.8" ILI9341 breakout board
- **Library version**: TFT_eSPI 1.4.10 (da2239f)
- **Calibration result**: `uint16_t calData[5] = { 485, 3382, 316, 3419, 3 };`

Thanks

Bodmer:
The intent was that the screen should be calibrated for the rotation you intend to use. The library does not handle dynamic rotation of the touch coordinates.

I do not plan to update the library at the moment.

ImpulseAdventure:
[...]

PR #376, opened 2019-06-12
Add note regarding setRotation() & calibration

ImpulseAdventure:
- Minor change to comments only
- Indicate that calibration must be updated if `setRotation()` is called with a new orientation
- Reference: https://github.com/Bodmer/TFT_eSPI/issues/365


Bodmer:
I am working on a user manual for the library so that will be the best place for the extra info. The ReadMe will then be simplified.

PR #389, opened 2019-07-04
Fix warnings in platformIO

luc-github:
Fix : warning: suggest parentheses around '-' inside '<<' [-Wparentheses]
Fix : warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
Fix : warning: pointer of type 'void *' used in arithmetic [-Wpointer-arith]


luc-github:
I close it as seems most of them are now fixed

PR #406, opened 2019-07-30
Fix: ST7735(160x80) setRotation(2) offset. ( test with M5Stick-C )

lovyan03:
When using setRotation(2), the drawing is off the screen by 26,1 pixels.


Bodmer:
I have tested this with a different board and it works with that one too so I have added the change.

Thanks for reporting this.

PR #410, opened 2019-08-08
Update README.md

Yul3n:


Bodmer:
I will be adding this information to the user manual and simplifying the readme as it has got rather long!

PR #411, opened 2019-08-08
Add const qualifier to initButton text

fcapano:
Pass the text as a pointer to const as the string is only copied


Bodmer:
In a future update it is likely the button text will be truncated to fit within the button so at I will not be making this change.

Thanks for your suggestion.

PR #419, opened 2019-08-28
I get a white screen only running the TFT_Print_Test and other examples

mepster:
I'm running an authentic Wemos D1 mini (ESP8266), and the Wemos TFT 2.4 Touch Shield ( https://wiki.wemos.cc/products:d1_mini_shields:tft_2.4_shield ) which is ILI9341 driver.

Trying all the TFT_eSPI examples, I get just a white screen on the display. (The same display unit does work with an ESP32, the Wemos D32 Pro.)

This fix makes the Print Test (and other examples such as Pong, and Cellular Automata, at 320x240) work on this hardware setup.

BTW my User_Setup.h has:
```
#ifdef ESP8266
  #define TFT_CS   D0  // TFT Chip select control pin (library pulls permanently low                            
  #define TFT_DC   D8  // TFT Data Command control pin - must use a pin in the range 0-31                       
  #define TFT_BL   -1  // TFT backlight control pin, aka TFT_LED                                                
  #define TFT_RST  -1  // TFT Reset pin, toggles on startup                                                     
  #define TOUCH_CS D3  // Touch Screen Chip Select, aka TS_CS                                                   
#endif
```

### DIFF
--- a/TFT_eSPI.h
+++ b/TFT_eSPI.h
@@ -128,7 +128,7 @@
   #define DC_C // No macro allocated so it generates no code
   #define DC_D // No macro allocated so it generates no code
 #else
-  #if defined (ESP8266) && (TFT_DC == 16)
+#if defined (ESP8266) && (TFT_DC == D0)
     #define DC_C digitalWrite(TFT_DC, LOW)
     #define DC_D digitalWrite(TFT_DC, HIGH)
   #elif defined (ESP32)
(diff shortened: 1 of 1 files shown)


akomelj:
I can confirm the exact same behaviour on my Wemos D1 Mini and Wemos TFT 2.4 Touch Shield.

All applications display white screen with library v1.4.20 and appear to work correctly after applying this patch.

akomelj:
> I can confirm the exact same behaviour on my Wemos D1 Mini and Wemos TFT 2.4 Touch Shield.
> 
> All applications display white screen with library v1.4.20 and appear to work correctly after applying this patch.

On the other hand, if I define
`#define TFT_CS   16`

instead of
`#define TFT_CS    D0`

or instead of officially recommended (PIN_* prefix)
`#define TFT_CS    PIN_D0 (which maps to D0)`

in my User_Setup.h, it works without patch as well. It appears D0 is not correctly mapped to numeric GPIO PIN value (16) in my configuration and versions of frameworks.

Bodmer:
Unfortunately this fix would break the libraries compatibility with platformIO users.

In the user setup file please use GPIO numbers, not D0 or PIN_D0.

This problem arises because of differences in pin mappings between ESP8266 board vendors.

I will think about other ways of dealing with this.

PR #437, opened 2019-10-03
Add LOAD_ALL_FONTS options to load all fonts

gordoste:
This is most useful to avoid passing all of the define's through platformio.ini


Bodmer:
Thanks for the suggestion but this is not something I wish to implement.

gordoste:
Fair enough. Thanks for answering and thanks for all the great work!

PR #453, opened 2019-10-29
Changed £ to $ sign for character 0x24 in 16p font

MiLeG:
The dollar sign in the Font16p.c file is defined to show the pound sign, instead of a dollar sign.
This pull request corrects just that


Bodmer:
Thanks, I have included as a configurable option to my master as I need the £ sign for one of my projects.

The change will be included in the next library release.

PR #475, opened 2019-11-21
Add: Destructor of TFT_eSprite.

lovyan03:
This prevents memory leaks if the user forgets to call deleteSprite.


Bodmer:
Agreed. Thanks,

PR #481, opened 2019-11-27
Add backlight LED pin output support

### LINKED ISSUE #477: ESP-32F Development Board with 1.44" 128x128 TFT LCD white screen
I have a [ESP-32F dev kit which came with the TFT LCD](https://www.aliexpress.com/item/33025684641.html) from Aliexpress

The docs pin as follow:
Screenshot from 2019-11-22 11-55-41

So I over-write the User_Setup.h setting as:

```
#define TFT_CS   26
#define TFT_DC   17
#define TFT_RST  16
#define TFT_MISO -1
#define TFT_MOSI 23
#define TFT_SCLK 18
#define TFT_BL 13
```

Even though the bag that came with the screen says it has ST7735 driver, but I can't get it to lit up.
Finally, I've tried all the listed driver, but only 
`#define ST7789_DRIVER`
I can see the white screen, but nothing else...
Screenshot from 2019-11-22 12-16-41

Can anyone suggest what I can try next? Many Thanks!

poky:
Fixes #477 ST7735 TFT_BL problem

### DIFF
--- a/TFT_Drivers/ST7735_Init.h
+++ b/TFT_Drivers/ST7735_Init.h
@@ -197,4 +197,9 @@
        }
        commandList(Rcmd3);
      }
+	#ifdef TFT_BL
+  	// Turn on the back-light LED
+  	digitalWrite(TFT_BL, HIGH);
+  	pinMode(TFT_BL, OUTPUT);
+	#endif
 }


Bodmer:
The back light drive capability has been added on an as needed basis for displays that support this feature. I think it is time to add this capability to all drivers, so I am postponing this update until I have thought about is further as once available for all displays then a function in TFT_E SPI to give brightness control is the logical next step.

The library is undergoing a major refactoring so all core enhancements to the repository are on hold except possibly bug fixes.

Bodmer:
Backlight support for all displays is included via:
```
#define TFT_BACKLIGHT_ON HIGH  // HIGH or LOW are options
#define TFT_BL   pin
```
I will not be adding PWM control as this is so easily added to the user sketch.

PR #495, opened 2019-12-13
Allow displays with CS/RST/DC lines connected indirectly

kostuch:
Hi,
At esp8266 platform there is not many I/O to free use.
In some projects it is necessary to add more with expander (popular PCF8574/PCF8575 or other).
I made some modifications to allow displays with control lines connected to such expander, so no digitalWrite() is possible.

Some remarks:
I2C is very slow comparing to SPI, even overclocked to 700kHz.
That's why using DC line over the I2C causes everything to slowdown and should be avoided. But sometimes I/Os are just exhausted and there is no other choice.
CS line is not so heavy used, so slowdown is not so noticeable, but if speed is crucial, then CS can be controlled by application/sketch. In this way programmer is responsible for activating CS, grouping graphic operations in batch and then release CS. This mode can be switched on when tft.begin is called.
RST line is toggled only once during initialisation, so there is no drawback.

It is also compatible with touchpanel and sdcard with CS lines connected to other expander outputs.

I tested it (on esp8266 with ili9341 module) with simple sketch and it works very well.


Bodmer:
Have you considered using overlap mode where the display shares the FLASH memory SPI port? This reduces the pin count needed to 2 (CS and DC. RST can be pulled high for some displays or connected to ESP8266 power on reset). The performance impact of the sharing is minimal.

My view is that it is better to keep the higher rate display connected to the ESP8266 and to offload sensors, ADC inputs and other digital I/O to a 3.3V Mini Pro which is linked to the ESP8266 with I2C or other serial interface means. There are so many libraries available for the Mini Pro that this becomes a relatively easy path to follow.

I think you have done a good job  but it is not an approach I think the library should adopt. I hope you are not too disappointed.

kostuch:
Hi,
Smaller modules (eg ESP-03) have no external pads exposing FLASH SPI, so sometimes it's physically impossible to use shared mode.
No problem with I/O when only TFT is used.
But when you have module with TFT + touchpanel + SDcard + ..., then demand for I/O lines grows.
Embedding additional MCU that must be also (re)programmed would be waste of resources and is not always possible because of power demand or size constraints.

Anyway, I'm not disappointed :)
I can use modified fork for my needs. Just offered a bit of flexibility for your great lib.

PR #503, opened 2019-12-28
Fixed some gpioMode issues

dracir9:
rtc_reg must be set to avoid some randomly occurring issue that caused my screens to freeze. It also caused some unexpected behavior with analogRead().
I added support for pins from 32 to 39, too.

### DIFF
--- a/TFT_eSPI.cpp
+++ b/TFT_eSPI.cpp
@@ -834,8 +834,23 @@ void busDir(uint32_t mask, uint8_t mode)
 // Set ESP32 GPIO pin to input or output
 void gpioMode(uint8_t gpio, uint8_t mode)
 {
-  if(mode == INPUT) GPIO.enable_w1tc = ((uint32_t)1 << gpio);
-  else GPIO.enable_w1ts = ((uint32_t)1 << gpio);
+  // vvvvvv Omitting this code may cause malfunction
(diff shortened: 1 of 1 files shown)


Bodmer:
The TFT_eSPI library does not use analogRead() anywhere so I assume there is a compatibility problem with a sketch or other library that just happens to be fixed by a change to the TFT_eSPI library and perhaps needs a change implemented elsewhere.

To understand why this change may be necessary I need to understand more about the hardware setup in which you are experiencing the TFT lockup.  I am guessing that you have a TFT in parallel mode and are trying to read a resistance touch screen, but this is just a guess and perhaps the analogRead is for a sensor?

dracir9:
Yea, you're right. I have a TFT in parallel mode and a resistive touch screen.
I finally managed to fix the issue outside the library. So if analogRead() is not used by the library and speed of gpioMode() is a priority I guess there is no reason for any change in TFT_eSPI.

Bodmer:
I have looked at the touch library and it sets the TFT_WR pin to be an analogue input. Since TFT_eSPI set TFT_CS permanently low for performance reasons any low analogue value on the TFT_WR pin would cause a spurious write tot he screen. I have changed the TFT_eSPI library to toggle the TFT_CS low only when a TFT write is in progress.

I suspect this is the true cause of the TFT lockup you experienced.

Thanks for reporting this.

PR #519, opened 2020-01-13
Fix json syntax

TioRuben:
Build fails when trying to compile with PlatformIO due to this malformed JSON


Bodmer:
Sorted, thanks.

PR #538, opened 2020-02-01
Add configurable drawButton function

justcallmekoko:
I wanted to be able to change datum, padding space, and label limit when drawing buttons

### DIFF
--- a/Extensions/Button.cpp
+++ b/Extensions/Button.cpp
@@ -86,40 +86,6 @@ void TFT_eSPI_Button::drawButton(bool inverted, String long_name) {
   _gfx->setTextDatum(tempdatum);
 }
 
-// Some added arguments for more custom buttons
-// Mod author: https://github.com/justcallmekoko
-/*
-void TFT_eSPI_Button::drawButton(uint8_t d, int padding, String button_name, boolean inverted) {
-  uint16_t fill, outline, text;
-
-  if(!inverted) {
-    fill    = _fillcolor;
-    outline = _outlinecolor;
-    text    = _textcolor;
-  } else {
-    fill    = _textcolor;
-    outline = _outlinecolor;
-    text    = _fillcolor;
-  }
-
-  uint8_t r = min(_w, _h) / 4; // Corner radius
-  _gfx->fillRoundRect(_x1, _y1, _w, _h, r, fill);
-  _gfx->drawRoundRect(_x1, _y1, _w, _h, r, outline);
-
-  _gfx->setTextColor(text);
-  _gfx->setTextSize(_textsize);
-
-  uint8_t tempdatum = _gfx->getTextDatum();
-  _gfx->setTextDatum(d);
-  
-  // Button label will be limited length unless string argument is provided
-  if (button_name == "")
-	_gfx->drawString(_label, _x1 + padding, _y1 + (_h/2));
-  else
-	_gfx->drawString(button_name, _x1 + padding, _y1 + (_h/2));
-  _gfx->setTextDatum(tempdatum);
-}*/
-
 bool TFT_eSPI_Button::contains(int16_t x, int16_t y) {
   return ((x >= _x1) && (x < (_x1 + _w)) &&
           (y >= _y1) && (y < (_y1 + _h)));
--- a/Extensions/Button.cpp
+++ b/Extensions/Button.cpp
@@ -61,6 +61,35 @@ void TFT_eSPI_Button::drawButton(bool inverted) {
   _gfx->setTextDatum(tempdatum);
 }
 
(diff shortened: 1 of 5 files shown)


Bodmer:
I understand that the objective of this change it is to be able to set the text datum plus an x position delta for the button label and so the label can be positioned more precisely within the button box. 

Assuming this is correct then I think a more generalised function is needed so that the y delta can also be set. There is already a -4 fudge factor in the [existing drawButton](https://github.com/Bodmer/TFT_eSPI/blob/master/Extensions/Button.cpp#L60) function which was supposed to be a temporary fix so I understand the problem you are looking to solve.  I think adding a new function to the button class is the way to go:
```
btn1.setLabelDatum(x_delta, y_delta, datum);
```
Then it is clear to the user they have the option of adjusting the label position and optionally the datum. This also avoids the situation that I find confuses noobs where identically named functions have different options.

The term "padding" is already used in the library to set the width to be blanked to background when rendering text hence the change to x_delta, y delta.

To let you know the direction of my thinking... the button class is rather old now (derived from a somewhat dated Adafruit_GFX library) and the processors that TFT_eSPI is intended for have the capability of creating more visually appealing buttons with colour gradients and shadows so a brand new GUI class in on my "to do" list, but the button class will be retained for backwards compatibility. 

I have experimented with LittlevGL which has an example for TFT_eSPI built into it but that is a very heavyweight library that consumes vast amounts of RAM. It is sophisticated and capable but also very noob unfriendly and needs an understanding of more advanced programming concepts to avoid development frustration.

justcallmekoko:
So I do have a couple of questions then. This is my first merge request ever so I will probably have a lot of questions.

So it sounds like there are a couple of options here.
Would a proper move be a supplemental `drawButton` function with a more distinctive name for the noobs as you put it with the option to adjust the y delta (replacing "padding" with delta)?

Or is the preferred action to create the separate `setLabelDatum` function to supplement the original `drawButton` function?

Bodmer:
setLabelDatum(x_delta, y_delta, datum); is the preferred approach, with the existing function adapted to accept the settings from that function.

A commented generic example (i.e. will work on any screen size) will also be needed so that users can see how the new function works. A single button in the middle of the screen with some example position adjustments demo'ed is fine, perhaps based on you own need for adustment.

I suggest the x and y deltas are mandatory (no defaults in the function prototype) and the datum is set to MC_DATUM. viz:
```
setLabelDatum( int16_t x_delta, int16_t y_delta, uint8_t datum = MC_DATUM ); 
```

justcallmekoko:
Awesome! Thanks for the guidance here. I will get back to you with something good.

Bodmer:
You will have to add 3 new [private global variables here](https://github.com/Bodmer/TFT_eSPI/blob/master/Extensions/Button.h#L29-L35).

For example, change to:
```
 private:
  TFT_eSPI *_gfx;
  int16_t  _x1, _y1, _xd, _yd; // Coordinates of top-left corner
  uint16_t _w, _h;
  uint8_t  _textsize, textdatum;
  uint16_t _outlinecolor, _fillcolor, _textcolor;
  char     _label[10];
```
Set to the global variables to defaults in the [button init](https://github.com/Bodmer/TFT_eSPI/blob/master/Extensions/Button.cpp#L21) function. They are then available to be changed by the new setLabelDatum function and used in the drawButton function.

justcallmekoko:
Got it. Thanks

justcallmekoko:
Well...it only took 8 commits for some reason but here it is

Bodmer:
Thanks.
The number of commits makes it hard to review and see the end result. Can you squash into a single commit for review. I just want to see a minimal set of changes that only affects the minimum set of files, e.g. no sequential changes to User_Setup.h. This minimises review time and lets me see just the required changes..
Also move the variable initialisation to the constructor, this will avoid needing to constrain a user to initialise the button before using setLabelDatum().

justcallmekoko:
I will squash the commits. Is there an option on github that allows me to do that?
I thought the "files changed" tab only shows the minimal changes.

I'm not sure what you mean by moving the variable initialization to the constructor. I currently have it set to initialize the _xd, _yd, and _textdatum in the initButtonUL which is called by the constructor.

Bodmer:
Each class has a constructor, for the button class [this is it](https://github.com/Bodmer/TFT_eSPI/blob/master/Extensions/Button.cpp#L5).

Not sure if you can do this from the GUI or web but [this may help](https://github.com/servo/servo/wiki/Beginner%27s-guide-to-rebasing-and-squashing).

justcallmekoko:
Wow I'm stupid. I totally blanked on what a constructor is. I will get that corrected and I will squash the commits and get back to you. Thanks for being patient.

PR #541, opened 2020-02-06
Fixes button text alignment and added keyword

justcallmekoko:


Bodmer:
I have merged. There are a few other tweaks needed like raising the issue number but I will do that.

justcallmekoko:
I'm glad I was able to contribute. I have learned some important lessons for the next time I want to do something like this.

PR #543, opened 2020-02-11
update: pushRotated tweak.

lovyan03:
Increased speed when using transparent color.

Improved pushRotated accuracy.
( Probably enough accuracy is obtained without using pushRotatedHP. )

Changed scanning direction to horizontal.  
( I think that the pushRotated from sprite to sprite can be sped up based on this algorithm. )


Bodmer:
Thanks, tests on a rotating rectangle show it is indeed a high precision and faster method. I have created the equivalent Sprite to Sprite function so will add that later.

PR #584, opened 2020-03-27
Added Second SPI selection option for Touch-pad driver on STM32 and ESP32

HamidSaffari:
I tested in both Parallel and SPI LCD transactions and in default SPI or second one with no problem.
It covers and resolved these problems:
1.Touch was disabled when using parallel mode.
2.Speed up the Main LCD transaction by not interfering with Main SPI.
3.Do not mess with occupied pins in parallel mode which is 8 or 16 pins that better be in continuous order when one of the SPI pins is in between.


Bodmer:
Thanks, but this is not a feature that I want to add. The touch support included is intended to be minimal, basic plus simple to wire and configure for noobs. This more advanced feature would be better added in a separate Touch library [like this one](https://github.com/achillhasler/TFT_eTouch).

HamidSaffari:
It's like before when you uncomment  #define SPI2_for_TOUCH_PORT  in user_setup. also the number of wiring in both cases is the same.
Beside it is not possible when using TFT in parallel mode to activate touch-pad because in your code when using parallel connection the SPI is disabled.
And for those reason GUIslice did that too. https://github.com/ImpulseAdventure/GUIslice

Bodmer:
I understand but I still believe this belongs in a support library that for example also handles the superiorcapacitance touch driver chips. Eventually the touch handler in TFT_eSPI will be deprecated when such a library is available.

PR #595, opened 2020-04-10
removed unused variable ly from fillCircle()

cyberman54:
removed unused variable (ly = y0) to sanitize compiler warning


Bodmer:
The circle fill algorithm has been updated and the variable deleted.

PR #603, opened 2020-04-19
Master

adamczuk-piotr:
Is there any reason that destructor is not virtual in TFT_eSPI class?


Bodmer:
The base class is not designed to be deleted as it has grabbed hardware resources and these would need to be released gracefully. In practice there should be no need to call the TFT_eSPI destructor or use polymorphic pointers from the base class to the derived Sprite class.

PR #607, opened 2020-04-27
Fix: Font7srle  Character '1' is missing two dots.

lovyan03:
When I drew character '1' over character '8', I found two dots short.

IMG_2356

Fixed image:
IMG_2354


Bodmer:
Thanks!

Issue #632, opened 2020-05-17
functions not in scope


LongLiveCHIEF:
I've got a fresh platformio project, and the first thing I've added is TFT_eSPI.  My `main.cpp` is a direct copy/paste of the `READ_USER_SETUP` example.

When I build i get `was not declared in this scope` errors for `printProcessorName`, `getPinName` and `lationn`.

When I try the `TFT_meters` example, i get even more `was not declared in this scope` errors for just about every function used.

Here's a link to this first commit, and when built will reproduce errors: https://github.com/LongLiveCHIEF/benchtop_psu/tree/e6a92bcce8809b8479249a72282ca599e01ef185

I've also tried it using `-include` build flag instead of the `-D` vars, with the same result.

LongLiveCHIEF:
I'm starting to think this is a platformio error with how it's executing the gcc compiler.

LongLiveCHIEF:
n/m... it's a forward declaration thing.  Fixed with reordering definitions.

Bodmer:
If you let me know which order worked then I will update the example.

PR #637, opened 2020-05-20
make TFT_Meters cpp compatible

LongLiveCHIEF:
When copy/pasting example to a platformio project, you get compiler errors with the `TFT_Meters` example due to missing declarations, (processing language hoists declarations, C++ does not)

### DIFF
--- a/examples/320 x 240/TFT_Meters/TFT_Meters.ino
+++ b/examples/320 x 240/TFT_Meters/TFT_Meters.ino
@@ -31,50 +31,62 @@ int value[6] = {0, 0, 0, 0, 0, 0};
 int old_value[6] = { -1, -1, -1, -1, -1, -1};
 int d = 0;
 
-void setup(void) {
-  tft.init();
-  tft.setRotation(0);
-  Serial.begin(57600); // For debug
-  tft.fillScreen(TFT_BLACK);
+// #########################################################################
+// Update needle position
+// This function is blocking while needle moves, time depends on ms_delay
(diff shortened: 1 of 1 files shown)


LongLiveCHIEF:
related #632

Bodmer:
I would think this applies to many more examples?  My understanding is that PlatformIO when configured for the Arduino environment handles the lack of early declarations.

LongLiveCHIEF:
@Bodmer platformio just calls the compiler, and to my knowledge doesn't do any handling of declarations unless specified.

I wasn't entirely sure, so I did do a little digging, and early in their docs they have this section on [Converting Arduinio file to C++ manually](https://docs.platformio.org/en/latest/faq.html#convert-arduino-file-to-c-manually)

I thought it was strange that they used the word "manually" implying that there is an automatic feature somewhere... but I haven't been able to find it yet.

My process when installing libraries with PIO has been to declare them in the `platformio.ini`, and then go to the libraries tab and pick and choose an example and copy/paste it into one of my `src/*.cpp` files to test it. 

If I'm not missing something, that means that yes this problem would apply to virtually all of the examples in this library at the moment.

If you want, I can do more digging to confirm.

Bodmer:
The library is targetted at the Arduino environment. The link to PlatformIO has been added on request of a user.

I suspect that users of PlatformIO are used to making declaration changes to examples to get them to compile. At least it seems that the main library does not create such errors!

I will mark as a potential future enhancement.

PR #692, opened 2020-07-11
Updated to match the available Class Methods

toetoast:
Placing all the TFT Class and Sprite Class methods here so that the IDE will show them in a different colour:

image

### DIFF
--- a/keywords.txt
+++ b/keywords.txt
@@ -138,11 +138,11 @@ getUnicodeIndex	KEYWORD2
 decodeUTF8	KEYWORD2
 drawGlyph	KEYWORD2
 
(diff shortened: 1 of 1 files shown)


Bodmer:
My local master has been updated already and will be pushed to Github soon. Thanks for the suggestion.

toetoast:
No worries. I've got some new enhancements for your consideration soon. I'm just new to GitHub as we use BitBucket in work, so I'm learning fast! Cheers Tony

Bodmer:
OK, it is best to raise an issue to see if a pull request will be accepted on a particular topic as their are assorted updates in the development pipeline.

PR #707, opened 2020-07-25
a few fixes for STM32 parallel bus operation

zerog2k:
- fix error when using STM_PORTB_DATA_BUS
- dont setup read pin if not using TFT_RD (The device I'm porting using [Autool X50 Plus] has 8-bit parallel bus, but sadly, they tied RD to vcc and did not wire this up to mcu.)

Other than that, thanks for the great library... (Going to try to see if GUIslice will work good on top of this.)


Bodmer:
I will add this feature. I will code myself as the ESP32 and generic processor code is also affected for completeness and it must cope where the TFT_RD pin is defined as -1 (can be the case as for TFT_RST and the Arduino digitalWrite function then ignores the pin).

Issue #700, opened 2020-07-22
Add more virtual modifiers to existing functions in TFT_eSPI to allow for polymorphism.


dr-gino:
The TFT_SPIFFS_BMP sample has a nice drawBmp function that draws directly to the screen. I wanted to add an extra argument to the drawBmp function to allow drawing on either the screen or a sprite. 
e.g.:
```void drawBmp(TFT_eSPI &surf, const char *filename, int16_t x, int16_t y)```

My intention was to be able to pass either an instance of TFT_eSPI or TFT_eSprite to it. However the functions used (```getSwapBytes```, ```setSwapBytes```, ```pushImage```) don't have the ```virtual``` modifier set in ```TFT_eSPI``` that will allow for polymorphism to work. So when passing in a sprite it always draws on the screen instead.

Adding the ```virtual``` keyword to those function in the header file made it work. May I suggest making all functions that are overridden by TFT_eSprite, ```virtual``` in ```TFT_eSprite```? This allows for writing functions that can take either the screen or a sprite as the surface to draw on.

**Comment by Bodmer** [comment]
OK I can do that!

To show library users the possibilities, can you first create a well commented example demo sketch. Also it would be useful to list what other functions that you think could be usefully made virtual. Bear in mind this is an Arduino library and the more esoteric capabilities in C++ are likely to remain a mystery to most users so examples are the key to such features being used.

**Comment by dr-gino** [comment]
I'm not sure how much these changes warrant a new example since I noticed that in the ```TFT_eSPI``` class that a couple of member functions such as ```drawPixel(), drawChar(), height(), width()``` etc. were already marked virtual, it only makes sense to extend it to other functions. Here's a list of functions in the TFT_eSPI that should be marked as virtual as well in the TFT_eSPI.h file:
```
setSwapBytes
getSwapBytes
pushImage (int32_t, int32_t, int32_t, int32_t, const uint16_t *)
pushImage (int32_t, int32_t, int32_t, int32_t, uint16_t *)
readPixel
setPivot
getPivotX
getPivotY
setRotation
getRotation
```

I slightly modified the `drawBMP()` function from the TFT_SPIFFS_BMP example to accept a surf argument. This allows you to use that same function for drawing directly on the screen but also pass in a TFT_eSprite. Much nicer than duplicating the function for both cases.

```
void drawBmp(TFT_eSPI &surf, const char *filename, int16_t x, int16_t y) {

[...]

PR #735, opened 2020-08-23
Add virtual modifier to more functions in TFT_eSPI class to allow for dynamic binding.

dr-gino:
Pull request to address issue #700.


Bodmer:
This increases code size and I suspect the majority of Arduino users will not benefit. The changes are quick to do by the users who wish to use polymorphism.

Issue #769, opened 2020-10-02
Drawing within a clipping region


ImpulseAdventure:
Hi @Bodmer -- fantastic work as always on the library!

I am wondering if there may be a way to render a BMP (eg. `pushImage()`), but only transfer a rectangular region of the overall image? In other words, I would like to perform pushImage, but constrained within a target clipping rectangle.

We already benefit from your optimized rendering in the GUIslice GUI, but such a feature would enable fast rendering during partial screen redraws (when buttons and visual controls are updating), as well as benefit the general TFT_eSPI user-base who may desire even faster updates.

I had originally thought that **setAddrWindow** / **setWindow** could be used, but of course these are already leveraged during row rasterization. Providing an offset to the top-left corner of the clipping region into the `data` param won't properly constrain `dw`, etc. Is there an alternate method for users to define a clipping region without leveraging offscreen/sprites?

Note that I have locally created a modification to TFT_eSPI that adds a **setClipRect()** API which can constrain the ensuing calls to pushImage() (and optionally other APIs where you are currently clipping on the screen boundaries). Seems to work quite well, in case you are interested in a PR.

Thanks!

Bodmer:
If the image is stored in a Sprite then there is a [function to render a window of the Sprite to the TFT](https://github.com/Bodmer/TFT_eSPI/blob/master/Extensions/Sprite.cpp#L792-L796). This works with 16,8,4 and 1 bit per pixel images.

See #712

ImpulseAdventure:
Hi @Bodmer -- thanks for pointing out a possible approach!
Yes, I did note the option to use sprites, but my hope was that the clipping would not depend on the use of an offscreen bitmap.

IMHO, it looks like it should be relatively straightforward to add this capability with only minimal changes.
If you are interested in a pull request for this feature, I have submitted one for consideration in #770 .

Bodmer:
OK, I thought you were looking at just image cropping. I have a code branch that implements TFT display viewports with full cropping in all functions but this has not been fully tested. This was created so the TFT datum can optionally be moved to the viewport corner for menus,   virtual screens etc, but if it the datum is not moved it would give the functionality you ask for.

Bodmer:
I have created a "Viewport" branch that includes the viewport capability. I have included an option to allow the coordinate datum to remain at the top left corner of the screen. By default it relocates to the top left corner of the viewport.

[...]

PR #770, opened 2020-10-03
Add support for graphics clipping

ImpulseAdventure:
# Drawing inside clipping regions
This enhancement adds `setClipRect()` which defines the clipping region used for the drawing primitives that follow (per #769 ).
- `setClipRect(int32_t x, int32_t y, int32_t w, int32_t h)`
- This can be useful when only a portion of an image needs to be updated; speeding up the redraw in the process

## Supported functions
I tried to update all of the APIs that previously performed screen region clipping to support the custom clipping region. This means that the following APIs now support the clipping region:
- `pushImage()`: all variants
- `drawChar()`: the default function
- `drawPixel()`
- `drawFastVLine()`
- `drawFastHLine()`
- `fillRect()`

### Functions not updated
- Sprite operations were not touched
- DMA operations were not touched
- I have not touched the `drawChar()` UNICODE variant

## Notes
- The clipping region works with dynamic `setRotation()`, and is reset accordingly
- I would expect this enhancement to have no negative performance impact (it should only improve performance)
- If a user has set a custom clipping region, it can be reset to the default by assigning the entire display: `tft.setClipRect(0,0,tft.width(),tft.height());`

## Testing
- Testing of this revision on a number of example sketches so far appears to show it to be working.
- Attached is a modified version of **TFT_graphicstest_one_lib** that demonstrates the result of a custom clipping region. You can un/comment the TEST_CLIP macro to observe the changes.
- [TFT_graphicstest_one_lib-clip2.zip](https://github.com/Bodmer/TFT_eSPI/files/5321055/TFT_graphicstest_one_lib-clip2.zip)

Thanks for your time

### DIFF
--- a/TFT_eSPI.cpp
+++ b/TFT_eSPI.cpp
@@ -909,7 +909,7 @@ void TFT_eSPI::pushRect(int32_t x, int32_t y, int32_t w, int32_t h, uint16_t *da
 void TFT_eSPI::pushImage(int32_t x, int32_t y, int32_t w, int32_t h, uint16_t *data)
 {
 
-  if ((x >= _clipX1) || (y >= _clipY1)) return;
+  if ((x >= _clipX1+1) || (y >= _clipY1+1)) return;
 
   int32_t dx = 0;
   int32_t dy = 0;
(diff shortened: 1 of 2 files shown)


Bodmer:
See "viewport" branch of TFT_eSPI.

PR #800, opened 2020-10-23
Added New Custom font files

viveky1794:
I have added two files of custom font under /Fonts/Custom folder.
I am also Planning to add other Fonts along with some good Examples with proper explanation.
Please let me know if you find any issue.
you can reach me viveky.ece16@gmail.com


Bodmer:
Thanks for your input, however I think there are already enough examples and fonts in the library. The assumption is that users will create their own fonts in the style and size they desire.

I think the best way forward is for you to start a new repository.

PR #838, opened 2020-11-27
Added UserSetup File for Waveshare 125mhz SPI Rpi Display on ESP32

Tysonpower:


Bodmer:
Thanks, I will add the Waveshare display to the list of those supported.
An update that includes this is already in the pipeline.

PR #896, opened 2020-12-31
Added alphablending support to sprite

DavyLandman:
I noticed alphablending wasn't working even after setting the callback when it was against a sprite instead of the bare tft object.

Since sprites keep track of the current state of the pixels, we can just use `readPixel` instead of the callback for this.


Bodmer:
Thanks, this is a good idea so I have implemented it in the latest version. When a text background colour is NOT specified, then the rendering will use the sprite pixel colour for the anti-aliasing blending.

DavyLandman:
Cool, I didn't want to just ask for a feature if I could implement at least a suggestion myself.

A bit of a pitty that the sprite code has to repeat the logic of the smooth font extension.

But I really love the sprite abstraction 👍

PR #939, opened 2021-01-17
Fix uncompilable TFT_eSPI_ESP32.h when TFT_RD >= 32

maxgerhardt:
See diff. This does not compile otherwise.


Bodmer:
Thanks for finding this.

PR #957, opened 2021-01-24
fix small typo in README.md

frankalicious:


Bodmer:
There were quite a few typos so I have corrected them maually, think I've got them all in the ReadMe now. Thanks.

If you need help using the library then there is a new Discussion tab.

PR #1002, opened 2021-02-12
Added compiler switch SMOOTH_FONT_OVERRIDE_BG

barn53:
Hi,

today I started a discussion: https://github.com/Bodmer/TFT_eSPI/discussions/1001 about that topic.
Then I looked into the code and came up with this simple solution, which takes the balance between memory consumption (sprites) and flicker (blank out complete rectangle before drawing new string).

For me, this approach works great, as there are no leftover artefacts when overriding with smooth font, nor is there any flicker to see. Performance is also acceptable, at least for my project.

I decided for a compiler switch, but also a flag would be possible.

While  testing, I also encountered a strange situation:
With an unicode special space character (not 0x20) with no pixels in it e.g. Six-Per-Em Space U+2006 there are invalid values. With NotoSans Regular, size 50, I got the following values for U+2006:
ascent: 38,
maxAscent: 38,
maxDescent: 12,
descent: 12
gxAdvance: 8, 
gdX: 4294967246
gHeight: 1, 
gdY: 100

This is why I added the strange conditions to the code...

What do you think?

Regards,
Gerd

### DIFF
--- a/Extensions/Smooth_font.cpp
+++ b/Extensions/Smooth_font.cpp
@@ -1,6 +1,6 @@
  // Coded by Bodmer 10/2/18, see license in root directory.
  // This is part of the TFT_eSPI class and is associated with anti-aliased font functions
- 
+
 
 ////////////////////////////////////////////////////////////////////////////////////////
 // New anti-aliased (smoothed) font functions added below
@@ -365,7 +365,13 @@ void TFT_eSPI::drawGlyph(uint16_t code)
   if (code < 0x21)
   {
     if (code == 0x20) {
-      //if (fg!=bg) fillRect(cursor_x, cursor_y, gFont.spaceWidth, gFont.yAdvance, bg);
+
+ #ifdef SMOOTH_FONT_OVERRIDE_BG
+
+      if (fg!=bg) fillRect(cursor_x, cursor_y, gFont.spaceWidth, gFont.yAdvance, bg);
+
+#endif
+
       cursor_x += gFont.spaceWidth;
       return;
     }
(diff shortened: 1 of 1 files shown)


Bodmer:
The approach taken as you have surmised is to over-write the background as a single block for the given string. This is designed to cope with glyphs/scripts from different languages and writing systems including Right To Left (RTL) writing schemes. Unfortunately the success of this approach is very dependant on the font, the font metrics and the way chanracters overlap. Take this font for example:

fig

The situation here is quite complex due to the overlapping bounding boxes. I think the suggested approach would have a lot of "caveat" conditions attached, which in turn will lead to issues being raised if these are not fully explained. In practice the situation gets even more complex for some fonts in other languages.

If the situation is that the majority of font cases are OK, then using the setAttribute() function for the switch is a better approach, it can then be turned on/off by the sketch. For example turn on only for numberical values.

I am pleased you have developed a solution but without testing on a range of fonts I suspect it will only work in particular conditions. If you can create some examples that show this works well in the majority of cases including script  then I will look at this in more detail.

Bodmer:
I think a better approach would be to track the righthand y coordinate of the last rendered characters bounding box. For example the background of the text below can be cleared with 3 non-overlapping rectangles using that approach:

fig2

This approach can then also be used with RTL rendering.

Bodmer:
I tried a minimised rectangle clear with tracking of the righthand edge and saw minimal improvement in flicker.

I have scrapped this idea and decided to try a scanline buffer approach with a mask. This will eliminate flicker, minimise buffer data size (compared to using a character sized sprite) and behave more like the RLE fonts.

I have added this to my TODO list.

PR #1006, opened 2021-02-14
ESP32 8-bit parallell DMA support

Satnet:
I don’t know if it’s possible. But I very much like this library by Bodmer. Adding 8/16 bit DMA support would be amazing.


Bodmer:
Pull requests are used to actually add new features or correct bugs.

Since this is a request please raise in the [discussuions tab](https://github.com/Bodmer/TFT_eSPI/discussions).

PR #1048, opened 2021-03-05
fix typo in examples

frankalicious:


Bodmer:
Corrected, also mispelt "editing" in all those files and "original" in some files. Thanks.

PR #1081, opened 2021-03-17
Update ST7789_Init.h

Bluebrain2000:
added ST7789_RAMCTRL setting
otherwise the default value of the second parameter, 0xF0, causes LSB of green being copied to LSB of red and blue in 16 Bit color mode, making every uneven green value grayish

### DIFF
--- a/TFT_Drivers/ST7789_Init.h
+++ b/TFT_Drivers/ST7789_Init.h
@@ -25,6 +25,9 @@
   writedata(0x55);
   delay(10);
 
+  writecommand(ST7789_RAMCTRL);
+  writedata(0x00);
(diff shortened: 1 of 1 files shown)


Bodmer:
I tried this on a TTGO TM board which supposedly has a ST7789 and it has no effect.

I have tried it on a TTGO T display and it does improve the green intensity gradient.

Since this appears to have a benefit to some ST7789 displays and does no harm to others I think it is worth adding. I prefer the EF = 10 setting as this is then consistent with the 16 bit to 18 bit colour conversion built into the library.  So I have added this to the init file.

Thanks for reporting this and finding the solution.

PR #1153, opened 2021-04-25
Added device support for the WT32-SC01

nuvious:
Added support for the [WT32-SC01](https://www.seeedstudio.com/ESP32-Development-board-WT32-SC01-p-4735.html); code sourced and modified slightly from [seaniefs's example](https://github.com/seaniefs/WT32-SC01-Exp). Added support for using standard tft.init() by defining TFT_BACKLIGHT_ON.

### DIFF
--- a/User_Setup_Select.h
+++ b/User_Setup_Select.h
@@ -86,6 +86,8 @@
 
 //#include <User_Setups/Setup200_GC9A01.h>           // Setup file for ESP32 and GC9A01 240 x 240 TFT
 
+//#include <User_Setups/Setup201_WT32_SC01.h>
+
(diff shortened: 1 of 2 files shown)


nuvious:
P.S. Tested on all non-touch 480 x 320 examples. All of them work when the Setup201_WT32_SC01.h include is commented in (and the User_Setup.h is commented out).

Bodmer:
Thanks I have merged and then edited for consistency with other setup formats.

SDA/SCL lines have been deleted as these will not be managed by the library.

I suspect the [new library here](https://github.com/Bodmer/Touch_FT5x06) will work with the display touch screen and gives reduced touch position glitches compared to the Adafruit library.

PR #1154, opened 2021-04-26
Add support for SSD1351! (128*128 OLED)

arduino12:
Hi @Bodmer !

**Great work you did here :)**
This repo helped me yesterday to use my ST7789 240x240 LCD,
So in order to make it complete - I have added and **tested** my new SSD1351 128*128 OLED I got today..!

Hope you can approve this PR to the master so others can use this display as well :)

**PS:**
The code is a bit messy - I think it will be better to make a `class` for each display-
This way each driver can implement its own `init, setRotation, drawPixel, setWindow, invertDisplay` ....
And the users can use more than 1 type of display (like now I can't use ST7789 + SSD1351)..

Arad


Bodmer:
Thanks for your pull request. I do not have one of those displays so I am unable to test.

A few minor changes are needed, the tft_Write_16C/D must be eliminated as these macros have already proliferated more than is desirable )to take advantage of processor related performance benefits of 32 bit transfers at high SPI rates). In the SSD1351 this is not needed as the display is so slow anyway (data sheet quotes 220ns i.e. ~ 4.5MHz max) so any optimisation is of not real benefit. This also means the hardware interface macros can be kept out of the display specific files.

An example setup file is of benefit to other users that gives the maximum workable frequency for the display and the recommended speed. I assume it has only been tested with a SPI interface. I expect the processor strobe timing will be far too fast for an 8 bit parallel interface so I suspect this interface option will have to be excluded.

Apart from that all looks good.

Are you able to test on ESP8266, STM32 and RP2040?

arduino12:
**Hi,**
Thanks :)

I got these **[displays](https://www.aliexpress.com/item/1005002051502238.html)**  - using my `ESP32 DevKit`,
but their PCB only support 4/3 wires SPI- so I can't test parallel interface - but as you said - few people will need it anyway..

I started with `20MHz SPI` with the original 30cm wires - but after some shielding it runs `40MHz SPI` (160fps DMA cube example)!
So maybe the `tft_Write_16C/D` still worth it? - I will test and let you know tomorrow :)

I can add an example - I am thinking of making a `128x128` folder that will run on all displays >128 pixels - ok?

I have an ESP8266 I can test with, other boards will take me longer to buy and test..

Last thing is that I use `PlatformIO`, but I find it exhausting to use almost any example -
Because the `.ino` file format doesn't care for functions order and prototypes - but the `main.cpp` in PlatformIO does!
So for each example - I have to change the functions order-
(mainly take the `setup` and `loop` to the end of the file- because they use function **below** them)
Do you know a way around it? maybe we can fix it in a different issue?
Anyway my new example code will support PlatformIO and switching boards and displays will be simple via platformio.ini :)

Bodmer:
Interesting, the data sheet quotes an unusually slow 220ns clock cycle time but looking again at the low/high minimums for the clock that is 20+20 = 40nS minimum, so 25MHz.   This latter figure makes sense based on your results.
For the "tft_Write_16C/D" it is just a case of using the existing tft_Write_16 and moving the shift+OR to the cpp file, this keep things a bit cleaner and the other processor macros should work without more macro patches.

Early library development tests with separate display and IO classes showed a significant performance penalty, probably down to an extra layer of vector tables that prevented inlining, but with more performant processors like the ESP32 that is less of an issue.

Bodmer:
For got to mention, I did not mean examples, I won't be adding a 128x128 set. The longer term aim is to have generic sketches, at least for the smaller sub 240x240 displays, that use the width and height values. The curent example partitioning is just historical from 8 bit AVR libraries developed years ago!

What I meant was a user setup header file for the pins and settings, plus any caveats (write only or read display as well with a tested lower clock speed)

arduino12:
Hi @Bodmer,

I hope I fixed everything you asked for in the commits above,
My display as well as the common breakout PCBs for this display doesn't breakout the `MISO` line-
So no need to implement these `read` functions (also I can't test them)..

**[Here is a demo video](https://photos.app.goo.gl/vKDbGVXScCJWAh8b7)** (got 50fps using DMA + SPI@20MHz on ESP32) - very cool but required almost 15 minutes of changes to the **[animated_eyes](https://github.com/Bodmer/TFT_eSPI/blob/master/examples/Generic/Animated_Eyes_2/Animated_Eyes_2.ino)** example to compile with PlatformIO...

Let me know if you want me to change more stuff before adding this display to the master :)

Bodmer:
Looks good.

Bodmer:
@arduino12  See #1653 where a user has problems with an SSD1351. Can you help?

PR #1168, opened 2021-05-04
Dangerous edge case for non-null-terminated string

bobwolff68:
strncpy() does not place a null terminator when the max characters is reached of '9' in this case. So, any buttons with 9 or more in the source will wind up with a potentially non-null-terminated _label.


Bodmer:
Thanks for reporting this. It is also worth reporting this on the Adafruit_GFX library as that is the source of that code segment.

I will update TFTeSPI.

PR #1216, opened 2021-05-27
Create Setup52_LilyPi_ST7796.h

c-logic:
It runs better with the LilyPi with ST7796.
No mirrored Screen and faster Update-Speed with 40MHz.


Bodmer:
Link lost...

PR #1247, opened 2021-07-01
FAST_UNICODE_INDEX - faster binary search for index from unicode code point

en-ot:
Hello, Bodmer!
I am using this great lib for making my mp3 player.
The font I need for display must contain wide unicode range: cyrillic, japanese, math symbols, etc, all that can be found in mp3 tags. The number of glyphs in the font is about 13800.
I have noticed that when I print glyphs with high codes, the speed is much slower, compared to base ascii range chars. 
The function that searches index from code point is simply looks for the given code iterating all glyphs of the font in the worst case.
The function that I offer here is binary search, which execution time is O(log(N)), is much faster in average than original, which execution time is O(N), where N = gFont.gCount.
I am using ESP32 + ST7789 (TTGO-TM) with 16MB of flash. For tests I use four printing methods from examples Font_Demo_x_Array. By my measurements the speed improvement is from 1...7% (on basic 95 glyphs font) to 110% (on 13K glyphs font). This result means that original function index searching time in my case (13K glyphs) is comparable to printing time itself.
The function only works with the fonts ordered by unicode code point, and that's why there is #define in User_Setup.h: for that strange cases when font is not ordered.

### DIFF
--- a/Extensions/Smooth_font.cpp
+++ b/Extensions/Smooth_font.cpp
@@ -338,6 +338,37 @@ uint32_t TFT_eSPI::readInt32(void)
 ** Function name:           getUnicodeIndex
 ** Description:             Get the font file index of a Unicode character
 *************************************************************************************x*/
+
+#ifdef FAST_UNICODE_INDEX
+bool TFT_eSPI::getUnicodeIndex(uint16_t unicode, uint16_t *index)
+{
+  int above = gFont.gCount;
+  int bottom = 0;
+  int i = above / 2;
+  do
+  {
+    uint16_t code = gUnicode[i];
+    if (code == unicode)
+    {
+      *index = i;
+      return true;
+    }
+    if (code > unicode)
+    {
+      above = i;
+    }
+    else
+    {
+      bottom = i+1;
+    }
+    i = (above+bottom)/2;
+  } while (above > bottom);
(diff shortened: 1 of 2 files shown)


en-ot:
updates later than initial (7d85157) are incompatible: font format changed

Bodmer:
Some pull links are broken. Not sure what your last message means...

en-ot:
Hello Bodmer!
  
  I don't know why links are broken, but this link still shows the changes in the first commit:
  https://github.com/Bodmer/TFT_eSPI/commit/7d85157c3c454a445d5bd1450791cb22e5815fe2
  I added this version (7d85157) long time ago, so maybe Your version is went far away from mine already, I didn't check.
  
  In later versions (44d8aa1, 8aae3fe) I added support for fonts in separate partition in spi flash. These versions are incompatible with your library, examples at least. You can read the description in README.md in https://github.com/en-ot/TFT_eSPI.
  You can use it as an idea how to decrease ram and cpu time usage when loading large smooth font. And even several different smooth fonts can be loaded on the fly (into separate sprites, or changed while printing), with no performance loss.
  Since then I've made some changes in this library, which I use in my mp3 player, but the library itself's repository is not actual now.

PR #1261, opened 2021-07-16
Add support for using 4bits mode smooth fonts

Lucy2003:
By default,the create_font create a 8 bits/pixel fonts,but now it can be
transform into 4 bits/pixel and used on ESP8266/ESP32.I have tested it on
ESP32,both SPIFFS and flash array.

Main modifications :
1. add readInt16() and readInt8() functions;
2. add a flag c4bpp to struct gFont.
3. adjust 4bits mode display.
4. add a tool made by pure C to convert font files.

The Smooth Font API is compatible with previous.Compressed font format is described in
tools/create_smooth_font/compress_font/readme.md

Signed-off-by: Lucy2003 <53362310+Lucy2003@users.noreply.github.com>

### DIFF
--- a/Extensions/Smooth_font.cpp
+++ b/Extensions/Smooth_font.cpp
@@ -475,7 +475,6 @@ void TFT_eSPI::drawGlyph(uint16_t code)
     int16_t  xs = cx;
     uint32_t dl = 0;
     uint8_t pixel = 0;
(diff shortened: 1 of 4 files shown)


Bodmer:
Pull link seems to be broken...

PR #1287, opened 2021-08-11
Add mirror option on setRotation() method for ST7789 display 

zivyuan:
A mirror option used to flip display content with x axis, this is useful for HUD like projects.

This modify only tested on my ST7789 display with size 240x240.

Here is the display i use:
display st7789 with no cs pin

Before mirror:
before mirror

After mirror:
after mirror


Bodmer:
Pull link is broken...

PR #1345, opened 2021-09-21
Add support for smooth fonts in 4bpp mode

Lucy2003:
Note: This pull is almost the same as #1261,repull due to broken pull link.

By default,the create_font create a 8 bits/pixel fonts,but now it can be transform into 4 bits/pixel and used on ESP8266/ESP32.I have tested it on ESP32,both SPIFFS and flash array.

Main modifications :

- added support for loading and displaying 4-bit mode smooth font.
- added a tool made by pure C to convert font files.

The Smooth Font API is compatible with previous.Compressed font format is described in
  tools/create_smooth_font/compress_font/readme.md

Signed-off-by: Lucy2003 <53362310+Lucy2003@users.noreply.github.com>

### DIFF
--- a/Extensions/Smooth_font.cpp
+++ b/Extensions/Smooth_font.cpp
@@ -126,7 +126,7 @@ void TFT_eSPI::loadFont(String fontName, bool flash)
   gFont.gArray   = (const uint8_t*)fontPtr;
 
   gFont.gCount   = (uint16_t)readInt32(); // glyph count in file
-                             readInt32(); // vlw encoder version - discard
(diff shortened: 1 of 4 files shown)


Bodmer:
The work you have done is very good. I assume this change has been made to reduce the font file by half. Is there any performance benefit?

I am reluctant to add this due to the process required to generate the font files. I will think on this further as I think Run Length Encoding (as used on other fonts in TFT_eSPI) would be a better solution.

Lucy2003:
Oh,I think the data to read from flash or sd card is reduced by half and the speed should increase a bit.

Yeah,if other fonts format can be added into the library,this change may not be very useful.If so,I will close this happily : )

PR #1356, opened 2021-10-07
Fix HX8375D rotation values

DaveBben:
This is for the Adafruit HX8357D driver
hx8357d

For the HX8357D, the current bits that are written to the MADCTL control register cause all of the text to be mirrored. This affects every rotation value.

tft.setRotation(0)

original_0

tft.setRotation(1)

original_1

tft.setRotation(2)

original_2

tft.setRotation(3)

original_3

If we use the same command values from the ILI9486 rotation, then rotated text displays like normal.

tft.setRotation(0)
fix_0

tft.setRotation(1)
fix_1

tft.setRotation(2)
fix_2

tft.setRotation(3)
fix_3

tft.setRotation(3)


Bodmer:
Thanks for this.

I do not have one of those displays so was unable to test!

PR #1377, opened 2021-10-22
Support Software SPI processor

jonesfu:
* some project can't use MCU SPI inteface because
  it will break other real-time interrupt service

  so that user software spi instead of spi bus


Bodmer:
Hi thanks for your pull request.

I understand the problem but I am not keen on adding software SPI. he problem is that some processors (STM32 etc) can toggle pins too fast, so then the interface fails.  I know it will be OK on an ESP32 as the I/O clock limits the toggle rate but other processors would ,not be guaranteed to work, leading to conditions being imposed.

PR #1384, opened 2021-10-26
Allow changing the display size via TFT_eSPI constructor

johalaoui:
Hello,
This patch allows to change the display size via `TFT_eSPI display(w, h);` for the LCD drivers HX8357D, ILI9481, ILI9488, R61581, RM68140. I used it to “simulate” a smaller sized display (with ILI9488). I don’t know if the usage of the variables `TFT_*` instead of `_init_*` was intentional or a mistake, but I hope the change of these allows for more flexibility.
signal-2021-11-21-184115


Bodmer:
I think this will cause problems for displays with offsets and this will then lead to issues being raised.

PR #1423, opened 2021-11-15
Resolution-independent Starfield example

daxliniere:
As someone new to coding, I can't believe I'm doing this, but here we go... :)

I have modified the Starfield sketch (found in the 320x240 examples folder) to work on 480x320 screen size, though I **_think_** it will work with any screen resolution if you adjust the **int resolution_x = 480;** and **int resolution_y = 320;** lines at the top of the sketch. Simply changing the XY size of the original worked, but the 'emission point' of the starfield wasn't in the centre of the screen.

[TFT_Starfield.ino.TXT](https://github.com/Bodmer/TFT_eSPI/files/7538231/TFT_Starfield.ino.TXT)

I hope this is helpful, but please be aware that's it's only been tested on 480x320 screen so should be given a run on another size screen first.

All the best,
Dax Liniere.


Bodmer:
Thanks. As examples it is up to the user to adapt. If I find time I ill update all resolution dependant examples.

PR #1446, opened 2021-11-24
Setup for the ESP32 S2 with Adafruit ST7789 display

sbonaime:
Setup for the ESP32 S2 with Adafruit ST7789 display


sbonaime:
> 

I don't know but you can try changing the pin in 
User_Setups/Setup71_ESP32_S2_ST7789.h

PR #1547, opened 2022-01-11
SPI should be reformatted in defined ways

kaidegit:
Some screen like some ST7789 based screen do not support SPI_MODE0, but the RP2040 part forced to format SPI to SPI_MODE0.


Bodmer:
Thanks for this. I will be making some changes to the library that will fix this.

Bodmer:
Fixed in latest version 2.4.27

Thanks for raising this issue.

PR #1587, opened 2022-01-28
Add user setup file for LilyGo T-Display RP2040

LeFauve:
This is just a User_Setup file for supporting the LilyGo T-Display RP2040.
Thanks :o)


Bodmer:
I have made a few minor changes.

1. It is expected that the user [comments out this line](https://github.com/Bodmer/TFT_eSPI/blob/master/User_Setup_Select.h#L22) to favour the selected setup. So there should be no need to edit the User_Setup.h file.
2. The PWM pin for the analogue (PWM) backlight function should be defined in the sketch.

LeFauve:
I didn't get that the User_Setup.h file shouldn't be included when using a custom one. 
Sorry about that, it does make sense when you think about it.
Thanks for merging the other changes!

PR #1591, opened 2022-01-28
User Setup for esp32-s2 touch_ILI9341

gabrielcsapo:
Trying to get the esp32-c3 to work I pulled out my esp32-s2 and got that working without any issues! I have added a user setup for others to be able to get off the ground running.

IMG_5623


Bodmer:
Thanks for this. Since this is nearly identical, I have updated setup 70 with an optional touch pin (does not have to be 34 or 10) and SPI clock rate. Most of the defined setups can be used with touch as an option so I do not want to create dedicated touch setups.

gabrielcsapo:
Sounds good, yeah maybe the readme could reflect this? There are a lot of settings that are sort of tucked away unless you need them. I could update the readme or something to make this more apparent?

PR #1602, opened 2022-02-02
added USE_FSPI_PORT flag to support more ESP32-S2 boards

Brezensalzer:
Hi,
I have added a new flag USE_FSPI_PORT to explicitely set the FSPI port and support more ESP32-S2 boards.
This change allowed me to use this library on an Adafruit QtPy ESP32-S2 which won't work otherwise.

Thank You for that great library!

### DIFF
--- a/Processors/TFT_eSPI_ESP32.c
+++ b/Processors/TFT_eSPI_ESP32.c
@@ -11,16 +11,18 @@
   #ifdef CONFIG_IDF_TARGET_ESP32
     #ifdef USE_HSPI_PORT
       SPIClass spi = SPIClass(HSPI);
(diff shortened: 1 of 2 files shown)


Bodmer:
FSPI should be the Arduino selected default if USE_HSPI_PORT is not defined. Are you finding that is not the case?

Brezensalzer:
It didn't work for my QtPy ESP32-S2 out of the box
* Without USE_HSPI_PORT it doesn't compile and complains about VSPI...
* With USE_HSPI_PORT it compiles and uploads, but the screen stays black (it uses the wrong pins)
* With the changes in my pull request everything works on my board

Bodmer:
OK, thanks.

PR #1613, opened 2022-02-05
Prevent TFT_eSprite::pushToSprite() to skips pixels when dealing with…

LeFauve:
… transparency.

The ox counter is incorrectly incremented in TFT_eSprite::pushToSprite(), which create issues with sprites containing transparency.
For example, if you create two full screen sized sprites, and draw the same grid on both of them with a different color, and use pushToSprite with transparency to push one of them on top of the other, you'd expect the two grid to superpose perfectly, but instead you get this:

Incorrect

The top grid is the blue one, and it seems to skip one pixel every time a transparent area is crossed.

The PR fixes this problem.


Bodmer:
Thanks.

PR #1681, opened 2022-03-04
Adding ST7789 240x280 display offsets

Fischkopppp:
Hi Bodmer, could you add the offsets for the 240x280 ST7789 1.69 inch displays?
They are:
case 0
        colstart = 0;
        rowstart = 20;
case 1
        colstart = 20;
        rowstart = 0;
case 2
        colstart = 0;
        rowstart = 20;
case 3
        colstart = 20;
        rowstart = 0;

Thank you for your great library :)


Bodmer:
These were recently added to the Github master, they are in the[ rotations code](https://github.com/Bodmer/TFT_eSPI/blob/master/TFT_Drivers/ST7789_Rotation.h#L13-L16).  To ensure thay are used you need to specify the screen size and invoke offsets in your setup file as follows:
```
TFT_WIDTH 240
TFT_HEIGHT 280

#define CGRAM_OFFSET      // Library will add offsets required
```
Screen sizes are always specified for the default (portrait) orientation of the screen. This is typically with the display connection ribbon at the bottom, but that is not always the case.

A complete  example setup file is[ here](https://github.com/Bodmer/TFT_eSPI/blob/master/User_Setups/Setup203_ST7789.h).

Thanks for the pull request. If the library update has not worked for you then let me know.

PR #1682, opened 2022-03-04
Allow generic driver to use a custom SPI port

Frogomeli:
I have a SeeedStudio Wio Terminal.
I'm using this lib in VS Code with PlatformIO but the lib
don't use the `SPICOM` variable when it's a generic device.

Steps to reproduce (using a Wio Terminal):
* Create a project using PlatformIO
* Add the TFT_eSPI lib to the project
* Add in the build flags these lines
```
build_flags = 
	-DUSER_SETUP_LOADED=1
	-DILI9341_DRIVER=1
	-DHASSPI=1
	-DSPICOM=LCD_SPI
	-DTFT_CS=LCD_SS_PIN
	-DTFT_DC=LCD_DC
	-DTFT_RST=LCD_RESET
	-DTFT_BL=LCD_BACKLIGHT
	-DTFT_BACKLIGHT_ON=HIGH
	-DTFT_BACKLINGT_V=2000
```
* Compile and transfer to the device
=> only backlight works


Bodmer:
To be clear are you saying the Wio Terminal works with that change or only the backlight works?

Frogomeli:
Without this commit only backlight works.
With this commit the screen and backlight works.

Frogomeli:
We can rename to another variable if you want.
I have used this variable(`SPICOM`) as it was used in the fork of the lib by Seeed Studio.
But the forked lib is not updated to the last version, so i use the original your lib.

Ref:
https://github.com/Seeed-Studio/Seeed_Arduino_LCD/blob/6bc2cc1f71d14ed4b5d9b4d8d86f8de267bbf3da/User_Setup.h#L143

Bodmer:
Thanks.

PR #1803, opened 2022-04-29
explicitly list ILI9342 as supported

belm0:
readers may not know that ILI9342 is almost identical to ILI9341


Bodmer:
Thanks, I have rolled this in with other updates needed.

PR #1804, opened 2022-04-29
user setup: reflect rename of M5Stack setup header

belm0:


Bodmer:
Thanks, have rolled this in with other relateed changes.

PR #1806, opened 2022-05-01
note 1 bpp in Sprite setColorDepth()

belm0:


Bodmer:
I see there are quite a few incremental updates that did not propagate into all the comments. I will do a code sweep prior to the next release and update them. I have been meaning to do this for some time so thanks for the prompt!

Bodmer:
Comments updated.

Issue #1328, opened 2021-09-08
M5Stack - TFT LED voltage too high!


Sarah-C:
Hi everyone!

I noticed when using this library the M5Stack #include setup is running the illumination LED at a much higher level than that advised by M5Stack engineers. 

I'd asked them previously about running the device at 3V3 rather than the 2V8 the M5Stack drivers use, for sunny conditions and got this reply back:

```
My email (10th August 2021):
Is 2.8 volts for the backlight a "Comfortable viewing brightness", 
and isn't the maximum safe voltage it can be run at?
And is the 3.3 volts allowed by SetLcdVoltage not recommended, 
due to the LED getting hot/shortening the white LED life of the screen?
Simply put - is it safe to run the backlight at 3.3 volts all the time, for use in sunny conditions?

------------------------------------

M5Stack (support) <support@m5stack.com>  (10th August 2021)
Hi Sarah,
No problem for short-time running in 3.3V. Let's say 1 or 2 hrs.
Not recommend for long-time running. Due to heat and life time concerns.
```

Therefore the M5Stack setup - the LED voltage should be reduced to their standard 2V8 to prevent over-heating. It might even invalidate their hardware guarantee.

m-menon:
I am not familiar with M5Stack but I guess this issue [here](https://github.com/Bodmer/TFT_eSPI/issues/226) should give you some pointers or ideas.

Sarah-C:
> I am not familiar with M5Stack but I guess this issue [here](https://github.com/Bodmer/TFT_eSPI/issues/226) should give you some pointers or ideas.

I've posted some code for people to use in their M5Stack sketches for "Core" and "Core 2" units.

Bodmer:
The TFT_eSPI library has NO built in support for back light LED brightness control. I suspect that is part of the M5Stack library and there should be reported in that repository.

Ideally M5Stack should design their products so that they cannot be damaged by the user software setting.

Sarah-C:
@Bodmer   The initialisation code of TFT_eSPI initialises the backlight to max brightness:
https://github.com/Bodmer/TFT_eSPI/blob/master/TFT_Drivers/ILI9341_Init.h

That is not a user software setting, but is the driver code for the screen that replaces the libraries M5Stack uses. That's why this issue is relevant. It even overrides the M5Stack library if included!

Therefore using this library with the M5Stack option within the #defines will damage it. 
[...]

PR #1815, opened 2022-05-07
user setup for M5Stack Core2

belm0:


User_Setups/Setup73_M5Stack_Core2.h
```diff
@@ -0,0 +1,27 @@
+#define USER_SETUP_ID 73
+
+#define ILI9341_DRIVER
```
belm0, line 3: Actually I'd like to use ILI9342, since this is a landscape device.  But it would require adjusting the mirror-image correction in ILI9341_Rotation.h.  Something like this:

```c++
    case 0:
#ifdef M5STACK
    #ifdef ILI9342_DRIVER
      writedata(TFT_MAD_COLOR_ORDER);
    #else
      writedata(TFT_MAD_MY | TFT_MAD_MV | TFT_MAD_COLOR_ORDER);
    #endif
#else
      writedata(TFT_MAD_MX | TFT_MAD_COLOR_ORDER);
#endif
      // ... (and similar for other rotation cases)
```

Happy to add that refinement, though things are usable as-is with ILI9341 and `setRotation(1)`.

belm0: Considering it more, using the landscape driver would cause compatibility annoyance because setRotation(0) would have a different meaning between Core Basic (which also has landscape display), Core2 on M5Stack's TFT_eSPI fork, and Core2 using the stock TFT_eSPI config.  So I'll leave it as is.

Bodmer:
The library does not support the M5Stack Core 2.

Due to a design fault the Core 2 sets the backlight LED current to a high value by default which overstresses the LEDs. An additional driver library is required to set the current to a lower value. This is not managed by TFT_eSPI and is a product design issue that could easily have been avoided.

belm0:
@Bodmer please reconsider

> Due to a design fault the Core 2 sets the backlight LED current to a high value by default which overstresses the LEDs

I believe it's a misunderstanding.  From my observation, no power is given to the LCD or backlight on boot.  It's necessary to explicitly write commands to the AXP to power these devices.

It's fair enough to say it's out of scope for the display library to manage the AXP.  However, it's perfectly valid to use M5Stack's library or code to initialize the AXP, and then use TFT_eSPI to drive the display.

This is exactly what I'm doing, and it works great-- I'd like to see others have access to the same setup.

Again, the backlight isn't driven on boot of the device, and TFT_eSPI does nothing to enable the backlight, so it's harmless to add this config.

Bodmer:
See #1328.

belm0:
I'm familiar with #1328.  The user may have been incorrectly using the M5Stack Core setup, which sets `TFT_BL`, on a Core2.  In contrast, this PR leaves `TFT_BL` unset.

I have the Core2 device in possession, and as I explained the backlight is not powered at boot.  So the user is responsible for running code to power it on correctly.  It's not related to TFT_eSPI.  We should still be able to use this library to drive the display.

Bodmer:
OK, that is good to know.  Users expect to be able to see the examples working once the correct setup is invoked. In the Core 2 case this will not happen because the backlight will not be switched on, and so I will end up getting issues raised! Bear in mind many users are following online video tutorials and web pages, so variations to the approach causes problems.

If you create a repository with both a setup file for the core 2 and a commected example that configures the backlight and displays something then I will include a link to it in the ReadMe. Alternatively post the setup, code and instructions in a Discussion topic to help users.

BG6TNB:
While I don't know why PR wasn't merged. But i created a minimal template project and highlighted the issues related to screen voltage in the sample code, so hopefully this will be helpful to others.

[M5Core2-TFT_eSPI-Template](https://github.com/BG6TNB/M5Core2-TFT_eSPI-Template)

PR #1816, opened 2022-05-07
user setup: increase SPI frequency on M5Stack Basic Core

belm0:
The proposed values have been used for years in M5Stack's embedded copy of TFT_eSPI-- should be safe.

https://github.com/m5stack/M5Stack/blob/16db1b0b14982e89cf0def79be5f1bf682ccfc17/src/utility/In_eSPI_Setup.h#L262-L266

### DIFF
--- a/User_Setups/Setup12_M5Stack_Basic_Core.h
+++ b/User_Setups/Setup12_M5Stack_Basic_Core.h
@@ -27,7 +27,7 @@
 #define SMOOTH_FONT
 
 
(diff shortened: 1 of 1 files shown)


Bodmer:
The frequencies are deliberately set to value I found were reliable in tests.
Note the rated maximum in the ILI9341 data sheet is only 10MHz. It is expected that the user adjusts the frequency to give the performance and reliability they find acceptable. Some ILI9341 chips work reliably at 80MHz but YMMV.

PR #1832, opened 2022-05-16
Image conversion to C array - Update panda_png.h

sonictl:
I spent much time for searching for [png to C array] tool. Finally found in you jpeg example. I hope to add the web tool url here.


Bodmer:
Thanks,
There are quite a few online tools, this is my favorite as it handles direct binary file copy to an array and also it can handle image conversion to different formats including 16 bit 565 and 8 bit 332:
https://notisrac.github.io/FileToCArray/

This is convenient as it avoids the need to edit the output.

The plan is to update all examples to use that tool since it then avoids using different tools to handle different formats.

Bodmer:
Example updated. Thanks for bringing this to my attention.

PR #1834, opened 2022-05-17
Update LittleFS_PNG_DMA.ino

sonictl:
add a little description for 'DMA'  (Direct Memory Access) to make it clear for noobs.


Bodmer:
The [ReadMe ](https://en.wikipedia.org/wiki/Direct_memory_access ) includes the abbreviation expansion but I can understand this can get missed, unread or forgotten. Since it is used in other examples I think the ReadMe is the best place for it. Eventually I will write a user manual, but that is low down on the TODO list.

PR #1838, opened 2022-05-21
Fix ESP-IDF header include issue

dracir9:
Fix for issue introduced in commits 338d56ca4229619206ae166f42e63a36b55cddf8 and a815d77ec4b0435a1f0d96181ec58e25a7cc54f3. This caused the configuration from ESP-IDF menu to not be loaded.

The file `TFT_config.h` must be included after `Arduino.h` and before `User_Setup_Select.h` otherwhise it will not properly read the configuration and create the corresponding `#defines`

### DIFF
--- a/TFT_eSPI.h
+++ b/TFT_eSPI.h
@@ -27,10 +27,6 @@
 ***************************************************************************************/
 
 //Standard support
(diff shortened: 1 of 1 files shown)


Bodmer:
Thanks fr your support.

PR #1893, opened 2022-06-30
Update TFT_eSPI_ESP32.h

hongthiet1994:
fix bug using tft parallel esp32 and uart  togerther

### DIFF
--- a/Processors/TFT_eSPI_ESP32.h
+++ b/Processors/TFT_eSPI_ESP32.h
@@ -357,7 +357,7 @@ SPI3_HOST = 2
     #define clr_mask (dir_mask); WR_L
   #elif (TFT_WR >= 0)
     // Data bits and the write line are cleared to 0 in one step (1.25x faster)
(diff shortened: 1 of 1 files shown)


hongthiet1994:
fix bug
TFT_eSPI_ESP32.h:353:22: error: expected ')' before numeric constant
TFT_eSPI_ESP32.h:353:22: error: expected initializer before numeric constant
TFT_eSPI_ESP32.h:353:22: error: expected ')' before numeric constant

when use tft parallel 8 bit together with uart freeRTOS ESP32

Bodmer:
Looks like a nameing clash specific to this case and the ESP32.

I will think on this further as there are other processor code blocks that use that same name.

Bodmer:
Please provide example code that demonstrates the problem. I am unable to reproduce this error.

Bodmer:
I have search and replaced the names throughout the library so you should find you do not get the error anymore. Thanks for raising this.

PR #1900, opened 2022-07-07
Fix dev-platform name for raspberrypi

ivankravets:


Bodmer:
Note: The library supports the RP2040 microcontroller only, not the Raspberry Pi A/B/4 Linux boards.

ivankravets:
Thanks for merging! Yes, see https://registry.platformio.org/platforms/platformio/raspberrypi

PR #1961, opened 2022-08-15
Fix Anti-Aliased drawing in parallel mode (#1960)

### LINKED ISSUE #1960: Anti-Aliasing not always working in 8-Bit parallel mode
**Problem:**
When using 8-Bit parallel mode to interface the ILI9341 (or in theory any other) driver, some anti aliased drawing methods (specifically the ones using the drawWedgeLine() method under the hood) don't work when not supplying a fixed background color but rather reading the color from the display.

This is due to an oversight in the readPixel() method which always sets the CS pin of the display to high (using the CS_H macro) after it is done reading the pixel in PARALLEL mode. In SPI mode, the inTransaction flag is saved and the write transaction restored after the read operation of the pixel has been completed. This also reselects the display by pulling the CS pin low.
In addition to that, in SPI mode the CS line and other status flags are controlled using the begin_tft_read() and end_tft_read() methods while the parallel implementation simply sets the CS value using the CS_L and CS_H macro, ignoring the other status flags.
(Have I missed a good reason for this implementation or was it simply an oversight?)

**Effects:** 
When trying to draw to the screen using the drawWedgeLine() method or methods that utilize drawWedgeLine(), only a few pixels will be drawn to the screen until the first edge pixel is reached where the previous color has to be read from the screen. Any subsequent operations until the method returns show no effect because the CS line will not be set low again after reading the pixel. Sometimes nothing might show up on the screen at all, when the first pixel to be drawn already requires an alpha blend.

**Fix:**
To fix the issue, I have modified the readPixel() method and created a pull request where I will explain what I have changed in more detail.

**Some more information about my setup:**
IDE: PlatformIO
TFT_eSPI library version: 2.4.75 (latest at time of writing)
Processor: ESP32 (WROOM32 on Generic ESP32 Dev Module)
TFT driver: ILI9341
Interface type: parallel

melektron:
Note: fixes #1960

As described in the issue, an oversight in the readPixel implementation prevents anti aliased drawing using the drawWedgeLine() method in parallel mode because the transaction state is not restored in the readPixel function when using parallel mode.

Changes:
 * Made parallel implementation of readPixel use the same methods to initialize the transfer and configure the CS pin as the SPI implementation (begin_tft_read() and end_tft_read()). This should work as these methods contain checks for parallel/SPI themselves, however I have only been able to confirm this with my specific hardware combo (described below)
 * Added the restoring functionality of the inTransaction flag and the transaction state to the parallel implementation which was previously only present in the SPI implementation.
 * Moved the duplicate code that is now present in both the parallel and SPI implementations out of the parallel/SPI #ifdef statement so it is less confusing to update in the future. In theory this should not have modified the SPI implementation functionality wise but I do not have any SPI displays to test this with so it should be tested for compilation/runtime errors before merging.

As I am not able to test every scenario this might be used in and I certainly don't know every possible scenario so it is very likely I have missed something that breaks some other code. I would be happy if someone notifies me or fixes the issue if one is found. 

My hardware:
8-Bit parallel ILI9341 driver connected to ESP32 (WROOM32 on Generic ESP32 Dev Module)

### DIFF
--- a/TFT_eSPI.cpp
+++ b/TFT_eSPI.cpp
@@ -1158,6 +1158,8 @@ uint16_t TFT_eSPI::readPixel(int32_t x0, int32_t y0)
   #endif
 */
 
+  CS_H;
+
   #ifdef TFT_SDA_READ
     end_SDA_Read();
   #endif
--- a/TFT_eSPI.cpp
+++ b/TFT_eSPI.cpp
@@ -1069,9 +1069,24 @@ uint16_t TFT_eSPI::readPixel(int32_t x0, int32_t y0)
   // Range checking
   if ((x0 < _vpX) || (y0 < _vpY) ||(x0 >= _vpW) || (y0 >= _vpH)) return 0;
 
-#if defined(TFT_PARALLEL_8_BIT) || defined(RP2040_PIO_INTERFACE)
+  // Temporary color variable because different driver ICs use different
+  // bit order in colors
+  uint16_t color = 0;
 
-  CS_L;
+  // This function can get called during anti-aliased font rendering
+  // so a transaction may be in progress. The old transaction state
+  // has to be saved so it can be restored before returning
+  bool wasInTransaction = inTransaction;
+  if (inTransaction)
+  {
+    inTransaction = false;
+    end_tft_write();
+  }
+
+  // set CS low and initialize SPI for reading if required
+  begin_tft_read();
+
+#if defined(TFT_PARALLEL_8_BIT) || defined(RP2040_PIO_INTERFACE)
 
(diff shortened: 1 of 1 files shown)


Bodmer:
I think this might cause issues in other setups so I will consider further.

Thanks for reporting this.

Bodmer:
I have fixed #1960 is a simpler way. It appears the other suggested changes were refactoring the code to make it cleaner. Let me know if those changes fixed something else.

melektron:
Indeed most of the changes are refactoring for cleaner code. I will check if the fix works with my setup as soon as possible but it looks promising.

PR #1973, opened 2022-08-21
Add user setup for TAMC Termod S3

TAMCTec:


TAMCTec:
I'm sorry to bother, but it's been 24  days.

Do I need something extra to add a new setup file for my device, or is it not welcome for third party setup files?

If there's something i can do, please let me know. Thank you.

Bodmer:
I cannot see any of these boards for sale so it is not clear to me how this is of benefit to other users.

Please provide a link to the vendors page.

PR #1975, opened 2022-08-21
ST7735: Add support for RobotLCD Arduino shield

boul51:
My LCD didn't work until I applied this modification (apparently, my controller version expects some different commands for the frame control)
As specified in the commit, I took the modified command from the adafruit library.
I'm not sure at all this is the best way to introduce this change, maybe just an #ifdef in the Rcmd1 declaration would be better.
Or maybe an entire redefinition of Rcmd1 for this controller version, but I find it a bit overkill, and would duplicate a lot of code.


Bodmer:
I have assumed this is no longer needed after merging #1974

PR #2126, opened 2022-10-31
Fix typo and remove unnecessary error in TFT_config.h

goodsoft:
1. RGB configuration was configured incorrectly, and caused re-definition warnings.
2. MISO pin is not always set, so shouldn't trigger a compilation error (see e.g. `Setup25_TTGO_T_DISPLAY.h`).


Bodmer:
Thanks, I don't use the ESP-IDF so your input is appreciated.

PR #2195, opened 2022-11-21
Enable HSPI configuration for ESP-IDF

supcik:
When TFT_eSPI is used as an ESP-IDF component, the files Kconfig and TFT_config.h are used to configure the device. However, the selection of the SPI port (VSPI or HSPI) seems to be missing. This Pull request addresses this issue.

See discussion #2191

### DIFF
--- a/Kconfig
+++ b/Kconfig
@@ -212,6 +212,21 @@ menu "TFT_eSPI"
 
     menu "Display SPI config"
         depends on !TFT_PARALLEL_8_BIT
+    
+        choice TFT_SPI_PORT
+            prompt "SPI port"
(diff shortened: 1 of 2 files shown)


Bodmer:
Thanks, I assume you tested this!

supcik:
> Thanks, I assume you tested this!

Yes. I tested that the `USE_HSPI_PORT` symbol was properly set when HSPI is selected in "menuconfig". Then I tested with VSPI and HSPI, and I was able to use the display with both SPIs.

supcik:
About testing, I did not see any "Action" configures for this repository. Are you using something else for the continuous integration?

PR #2296, opened 2023-01-04
Added T-DISPLAY-S3 i8080 support for more than 33 data pins

mmMicky:


mmMicky:
Since ESP32-S3 registers are only 32 bits.
Operation of pins above 33 requires the operation of another register. I added a macro definition as a switch to switch between pins above or below pin 33.
#define TFT_DATA_PIN_OFFSET_EN

This commit is known to have an open issue.
Including both high and low number pins on data pins does not work properly. This modification was tested with T-DISPLAY-S3 and works fine.

Bodmer:
Thanks, a few tweaks will be needed so I have merged into a new branch.

What testing on "real hardware" have you done?

Bodmer:
I have made some minor changes to the test branch here:

https://github.com/Bodmer/TFT_eSPI/tree/T-Display-S3

Can you test this on your display?

mmMicky:
I used the library in this link with the alphaBlend_Test example and the example in the link below
carry out testing.

https://github.com/Xinyuan-LilyGO/T-Display-S3/blob/main/example/tft/tft.ino#L21

works fine.

I suggest adding the following initialization commands to this model and using macros to control whether to enable or disable, sometimes it will make the screen display better.

https://github.com/Xinyuan-LilyGO/T-Display-S3/blob/f155cb3835dd8c9fadf6a65f4cbd613e26d881f9/example/tft/tft.ino#L21

Bodmer:
I have added the new init sequence, this can be invoked by inluding the following line in the setup file:
```
#define INIT_SEQUENCE_3    // Using this initialisation sequence improves the display image
```
As in the [modified setup file here](https://github.com/Bodmer/TFT_eSPI/blob/T-Display-S3/User_Setups/Setup206_LilyGo_T_Display_S3.h#L6).

Please test this to confirm it has been incorporated correctly.

mmMicky:
image

I just tested with the latest version. It is found that the value of the 0X3A register of some screens needs to be modified before it can be used normally during initialization. This register is used to determine the bit depth of the byte pixel. Please add the following statement here. to run normally.

  writecommand(0X3A);
  writedata(0X05);

image

Bodmer:
This has been added and meged into the library master. Thanks.

Issue #2301, opened 2023-01-04
tft.init() generates "gpio_set_level(226): GPIO output gpio_num error" for T-DISPLAY hardware


pabloandresm:
For hardware T-DISPLAY, when doing tft.init() generates the following error on serial:
E (228) gpio: gpio_set_level(226): GPIO output gpio_num error

Tracking down the problem, I've found that the Setup25_TTGO_T_Display.h defines "TOUCH_CS" to "-1". So when TFT_eSPI::initBus() is called, it executes "pinMode(TOUCH_CS, OUTPUT);" which generates that error at runtime.

The solution is simply to comment out the "#define TOUCH_CS            -1" in the Setup25_TTGO_T_Display.h file. It just should not be defined at all.

Bodmer:
Thanks, master library has been updated.

pabloandresm:
Thank you Bodmer.

May I suggest also inside the TFT_eSPI.cpp ::initBus() an
if (TOUCH_CS >=0) {
....
}

Because removing the define TOUCH_CS might work for T-DISPLAY, but I've seen that define in other hardware.

I've put this too in my pull request.

Cheers,

PR #2302, opened 2023-01-05
fix for warning - GPIO output gpio_num error

pabloandresm:
fix for issue #2301

### DIFF
--- a/TFT_eSPI.cpp
+++ b/TFT_eSPI.cpp
@@ -541,8 +541,10 @@ void TFT_eSPI::initBus(void) {
 
 // Configure chip select for touchscreen controller if present
 #ifdef TOUCH_CS
-  pinMode(TOUCH_CS, OUTPUT);
-  digitalWrite(TOUCH_CS, HIGH); // Chip select high (inactive)
+  if (TOUCH_CS >= 0) {
+    pinMode(TOUCH_CS, OUTPUT);
+    digitalWrite(TOUCH_CS, HIGH); // Chip select high (inactive)
+  }
 #endif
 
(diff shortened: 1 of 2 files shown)


Bodmer:
This problem is specific to the Espressif procesors as normally setting the pin to -1 in the Arduino environment means "don't allocate a pin" to the receiving function. Espressif removed this check in board package 2.0.0 (it appears a new software team took over) which breaks compatibility.

I will need to think on this further as there are other places where this could cause problems, for example:
https://github.com/Bodmer/TFT_eSPI/blob/master/Processors/TFT_eSPI_ESP32.h#L283

pabloandresm:
Ah ok. Understood.

 

As I saw a couple of lines below the same “if >=0” but for TFT_RST, I assumed it could also be done with TOUCH_CS.

 

My bad.

 

Thank you for your prompt replies.

Pablo

 

 

De: Bodmer ***@***.***> 
Enviado el: Friday, 6 January, 2023 11:30
Para: Bodmer/TFT_eSPI ***@***.***>
CC: Pablo Martikian ***@***.***>; Author ***@***.***>
Asunto: Re: [Bodmer/TFT_eSPI] fix for warning - GPIO output gpio_num error (PR #2302)

 

This problem is specific to the Espressif procesors as normally setting the pin to -1 in the Arduino environment means "don't allocate a pin" to the receiving function. Espressif removed this check in board package 2.0.0 (it appears a new software team took over) which breaks compatibility.

I will need to think on this further as there are other places where this could cause problems, for example:
https://github.com/Bodmer/TFT_eSPI/blob/master/Processors/TFT_eSPI_ESP32.h#L283

PR #2411, opened 2023-02-15
Add support for LilyGo T-Dongle S3

agrabren:
These are becoming popular with some Home Assistant groups and would be nice to see easily selectable


Bodmer:
Added, thanks.

PR #2423, opened 2023-02-19
Fix TFT_ring_meter.ino example build

drabaioli:
Fix TFT_ring_meter.ino example build: typo in the include file.


Bodmer:
Thanks, updated manually.

PR #2430, opened 2023-02-20
8 bit parallel fix for ESP-IDF

dracir9:
Fixed a missing define that prevented using 8-bit parallel bus on ESP-IDF.
Simplified the LCD interface selection (SPI/Parallel) and added the available pin ranges for ESP32-S3


Bodmer:
@dracir9  See #2435, thanks.

PR #2474, opened 2023-03-07
Adapt to a new round screen

Maxwelltoo:
Added setup and extension files.  I have tested that the display and touch functionality of the screen works correctly when I run this code.


Bodmer:
I am willing to add a setup file for that display. The touch interface code needs to be built into a new independant support library.

PR #2493, opened 2023-03-12
Modified Create_font.pde procesing sketch to enable automatic generation of a C headers

troisiemetype:
Hello,

I have just modified the processing sketch, so it can create a C array.

It's implemented that way : once the `.vlw` has been created, it's opened as an input file, and parsed. A `.h` is created, and each value read in the font file is formatted in C style. `<pgmspace.h>` include is also created, as well as the `const uint8_t` var name and `PROGMEM` keyword.

I've added an option after the `fontSize` setting to enable the creation of this file.
I've also added a side option to _not_ open the folder once the sketch has run, because I was tired of closing it each time, when generating several fonts.

I've moved the filename creation out of the `try` block, so it can be used after when creating the `.h` file.

The last thing I've modified is to comment out the `fontName` copied from the font list : on fonts that have space in their names, it creates a folder with space names, but it also makes the varname in the header file having spaces in it. I didn't found a function in Porcessing to remove spaces from a String. Plus it seems to me it's a good thing to have a var name as short as possible in the programs.

I've tested it on a few fonts, it seems to be ok. The output is identical to the online tool you give in the Smooth Font examples.

Please tell if there anything you see that should be modified.


Bodmer:
Thanks. It will be a few weeks before I can check the changes.

Issue #2533, opened 2023-03-27
White space is way too small for monospaced font


rebior:
1. Description
I created vlw font files from the font "Noto Sans Mono" on Linux. The space character is excluded as mentioned in the guide. I use drawString to put smooth text on the screen and all characters have equal width except for space characters, that are way smaller.

3. IDE: PlatformIO/Arduino
4. TFT_eSPI library version: 2.5.0
5. Board package esp32dev (cant't find exact version)
6. Procesor ESP32
7. TFT driver ST7735_BLACKTAB
8. Interface type SPI

The following screenshot shows the two strings ```"HOST: "``` and ```"  IP: "```, both left aligned at the same x position. I would expect for the colons to be vertically aligned (which is the case if there are only characters other than space)
![Screenshot](https://user-images.githubusercontent.com/80133983/228068274-b2364b53-6514-4a8b-b2e8-31f71b767edc.jpg)

I can also provide a vlw file if it helps.

Bodmer:
The vlw files don't specify a space with, so this is guessed from the font height here:
https://github.com/Bodmer/TFT_eSPI/blob/master/Extensions/Smooth_font.cpp#L139

In your use case gxadvance would probably work.

rebior:
> In your use case gxadvance would probably work.

This means I should set that value manually after loading the font?

rebior:
I created a pull request that provides a simple and backwards compatible solution for that issue: #2540

PR #2540, opened 2023-03-29
Optional hint for space width of smooth fonts

rebior:
Introduction of a SpaceHint enumeration that can
be used in the loadFont methods to affect the
space width of the font.

This is a simple solution for #2533


rebior:
Hello @Bodmer,
you didn't merge this PR into mast and write it needs consideration - do you have a decision already? Can a help with the consideration?

PR #2590, opened 2023-04-16
be able to customize without having to edit in the library directory

joseluu:
I am proposing a solution so that users do not have to modify the file in the arduino library area: User_Setup_Select.h

This makes it convenient when one has several hardware configuration for different projects.

how to use it:
The configurations of interest should be copied to the sketch directory
Sketch should be modified, for instance:
#include "Setup25_TTGO_T_Display.h"
#define USER_SETUP_LOADED 1
#include <TFT_eSPI.h>

Notes: 
unfortunately I could not use the construct in TFT_eSPI.h (line 58): #if __has_include(<tft_setup.h>)
somehow it does not work and I am not knowledgable enough in the arduino compilation process


mhightower83:
> unfortunately I could not use the construct in TFT_eSPI.h (line 58): #if __has_include(<tft_setup.h>)

With the more recent update to the Arduino ESP8266 core, something like this has worked for me.

In the Sketch folder create a file `SketchName.ino.globals.h`
```cpp
/*@create-file:build.opt@
// This is needed for TFT_eSPI to find the Hardware matching tft_setup.h file
// This must point to your Sketch folder
-I"/home/user/Arduino/SketchName/"
*/
```
Details about the global comment block can be read here:
https://github.com/esp8266/Arduino/blob/master/doc/faq/a06-global-build-options.rst

If you use Arduino ESP32, I think `build_opt.h` may work.

Bodmer:
See #2007.

The Arduino IDE blocks this approach. Attempts to bypass this  blockage can have unforseen and tricky to debug consequences. Considering it only takes a few moments to edit User_Setup_Select.h, that is the minimal effort and risk approach. Use PlatformIO for project based configuration in platformio.ini file.

PR #2657, opened 2023-05-16
Rename C -> CPP and make SPI.h required if needed.

Fabian-Schmidt:
Rename `Processors/*.c` files to `Processors/*.cpp` as they all contain C++ code.

`include <SPI.h>` only if display is spi display. For parallel displays the library is not needed.


Bodmer:
Are you having compile problems? If so, describe the software environment you are using.

Fabian-Schmidt:
I am using a ubuntu 22.04 environment for building. The library is used as part in an ESP Home project which uses platform io for build pipeline.
When using the C++ files with `.c` ending rebuilds and changes in files and defines where not detected properly. As soon as I named them cpp every was working and I was able to test the changes and continue changing the files.

Fabian-Schmidt:
Hi @Bodmer 
Do you think it is an issue renaming the Processors files from `.c` to `.cpp`?
As the files contain C++ code they should have the correct extension, even if the compiler does not complain.

Bodmer:
If I rename files as it will break various web links that refer to the files and content, so unfortunately this is a legacy naming situation that will have to remain. I can see that this will be inconvenient in your use case.

I would expect changes to be picked up correctly by the IDE/build environment. One solution may be to avoid the use of the precompiled code cache and use compile options to force a complete rebuild if it has been necessary alter the library code.

Bodmer:
If you can create a pull request with just the SPI mods in it then I will merge it.

Bodmer:
I see those SPI changes are trivial so I can edit manually. Thanks.

Fabian-Schmidt:
> If I rename files as it will break various web links that refer to the files and content, so unfortunately this is a legacy naming situation that will have to remain. I can see that this will be inconvenient in your use case.
> 
> I would expect changes to be picked up correctly by the IDE/build environment. One solution may be to avoid the use of the precompiled code cache and use compile options to force a complete rebuild if it has been necessary alter the library code.

Thanks for the explanation. That makes a lot of sense and I did not see this.

PR #2668, opened 2023-05-25
Make CONFIG_TFT_RST optional

oscarthorn:
Hi!

Thanks for maintaining such a great library!

I'm using esp-idf with a esp32-c3 and st7789 screen. With this setup the RST pin is optional. However in TFT_config.h not setting the reset pin throws an error. This PR changes that to throw a warning instead, and moves the CONFIG_DISABLE_WARNINGS setting to the top of the file so that you can disable the warnings also in TFT_config.h.

This brings the esp-idf compatibility more in line with the rest of the library to where RST pins are not always set.

### DIFF
--- a/TFT_config.h
+++ b/TFT_config.h
@@ -24,6 +24,14 @@
 #include "sdkconfig.h"
 
 /***************************************************************************************
+**                         Others
+***************************************************************************************/
+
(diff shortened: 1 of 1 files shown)


Bodmer:
Thank. I don't use the esp-idf so this is helpful. Thanks.

PR #2678, opened 2023-05-29
Fix warning unused-but-set-variable

mircopz:
Fix the following warning:
`variable 'swin' set but not used [-Wunused-but-set-variable]`


Bodmer:
Thanks for reporting this.

I have fixed this in a slightly different way to minimise ifdef clutter.

PR #2774, opened 2023-08-01
Added ST7796 (222x480) support

lewisxhe:
1. Added ST7796 (222x480) support
2. Added LilyGo T-PicoPro & T-BeamShield UsersSetups


Bodmer:
Thanks.

PR #2809, opened 2023-08-22
ESP32 Parallel DMA support

dracir9:
Added support for DMA transfers with 8-bit parallel bus.

Tested with ESP32-S3 and ILI9488 using Arduino IDE, PlatformIO and ESP-IDF. It should work with ESP32-S2 as well. All other displays using 8-bit bus should also be compatible unless they use a completely different protocol.

More testing is probably needed but I don't have more displays at my disposal.

LIMITATIONS: Using DMA with a parallel bus implies that all the bus GPIO are handled by hardware. Therefore, once DMA is started by calling initDMA() DO NOT use any non-DMA functions that write to the display until deInitDMA() function is called. Otherwise, it will cause undefined behaviour.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,15 @@
 
 idf_component_register(SRCS "TFT_eSPI.cpp"
                     INCLUDE_DIRS "."
-                    PRIV_REQUIRES arduino)
+                    PRIV_REQUIRES "arduino")
+
+function(maybe_add_component component_name)
+    idf_build_get_property(components BUILD_COMPONENTS)
+    message(STATUS "Components: ${components}")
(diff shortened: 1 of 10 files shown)


hemangjoshi37a:
here what does DMA means ?

hemangjoshi37a:
here what does DMA means ?

dracir9:
DMA stands for Direct Memory Access. It is a feature that enables data to be transferred without the intervention of the CPU. The goal is to let the DMA module transfer the color data from the ESP32 to the TFT without blocking your code. This results in faster execution and faster screen drawing.
The downside is that can be tricky to configure and generate confusing errors if not handled correctly.

PR #2911, opened 2023-10-22
Fixed issues with font 1 GLCD

JuniorPolegato:
Hi!

- I fixed font configuration via respective parameter in setCursor, calling setTextFont instead of just change textfont value:
`void TFT_eSPI::setCursor(int16_t x, int16_t y, uint8_t font)`
`{`
`  setTextFont(font);`
`  cursor_x = x;`
`  cursor_y = y;`
`}`

- I added the lost character of CP437 table into file glcdfont.c:
178 - 0xB2 - '`▓`' - unicode 9619
`0xAB, 0x55, 0xFE, 0x55, 0xAB,  // '▓' - 9619`

- I create a function for correct conversions between CP437 and uniCode, based on this https://opensource.apple.com/source/samba/samba-26/source/codepages/CP437.TXT.auto.html and similarity between 30 first bitmaps and utf-8 characters, and as there is no "`ã`" in CP437, I translated it to “`ä`”, also ☹ to 1, unicode doesn't have black, plus ☺ and ☻ 2:

`uint16_t TFT_eSPI::encodeCP437(uint16_t uniCode)`
`{`
`    switch (uniCode) {`
`   	 case 9785: return 1;`
`   	 case 9786: return 2;`
`   	 case 9787: return 2;`
`   	 case 9829: return 3;`
`   	 case 9830: return 4;`
`   	 [...]`
`   	 case 8730: return 251;`
`   	 case 8319: return 252;`
`   	 case 178: return 253;`
`   	 case 9632: return 254;`
`   	 case 160: return 255;`
`    }`
`    return uniCode;`
`}`

- Them added a call to it at block:
`#ifdef LOAD_GLCD`
`  if (textfont==1) {`
`  	cwidth =  6;`
`  	cheight = 8;`
`  	uniCode = encodeCP437(uniCode);`
`  }`
`#else`

Everything works fine, tested with my project:
[TTGOWeatherStation](https://github.com/JuniorPolegato/TTGOWeatherStation)


Bodmer:
I will think about this as I see the Adafruit_GFX library has incorporated a [CP437 work-around](https://github.com/adafruit/Adafruit-GFX-Library/blob/master/glcdfont.c#L99-L100) and I would prefer not to introduce new custom mappings. I have added a configuration switch but never got around to implementing the update.

JuniorPolegato:
Hi!
Cool, I also think it's better to use this work-around from the Adafruit_GFX library, but I think that whoever needs more characters has that freedom, so I created this [font editor](https://github.com/JuniorPolegato/Adafruit-GFX-font-editor) for an easy solution. 
Whatever, the conversion from UTF-8 to CP437 is necessary, as well as the setCursor fix, go ahead!

Bodmer:
Fixed in latest master. Not released yet.

PR #3032, opened 2023-12-11
Allow Custom Function to Write CS & DC to Minimize GPIO Usage

qzchenwl:
Add new init function to pass write_cs & write_dc callback, so that TFT_CS and TFT_DC can be connected to GPIO expander.


Bodmer:
Thanks for the contribution. However I think this adds a level of complexity for the average user that would not need. There is also the complexity of explaining usage and the likely Q and A support load. If you create a specific branch with implementation instructions then I can add a link to it.

PR #3068, opened 2023-12-27
Add TouchScreen support for displays with parallel bus.

piratfm:
Used internal ADC for touchscreens. Config avaliable for ESP32 Wemos D1 R32 with ILI9486. Requires additional wire modification as shown on https://github.com/s60sc/Adafruit_TouchScreen repo: (https://github.com/s60sc/Adafruit_TouchScreen/raw/master/extras/wiring.jpg)


Bodmer:
The ADC based resistive touch screens support would need to be added for all supported processors. In view of the variances between processors and screen configurations I think it would be complex to provide a "plug-and-play" solution without a lot of configuration options.

In view of this there are no plans to add this functionality and it is best implemented in specific third party libraries.  In the longer term the plan is to deprecate touch support in TFT_eSPI.

PR #3134, opened 2024-01-25
Cube mx

wvdakker:
This is a PR for using TFT_eSPI with the STM32CUBEIDE (including CubeMX) or Platformio with the stm32cube framework.
I have tested the patch with the parallel version of a ILI9481. 
To activate the use of the framework a '-D CONFIG_TFT_eSPI_STM32CUBE' must be used. Omitting the option falls back to the original use of TFT_eSPI.

There is a README in the Doc/CubeMX  directory which describes all steps.

For now it is working with the Parallel interface, I am working on the SPI interface which will be added in the next days.
However I made a PR to get a review so I can correct all issues during the next upload.

Have any questions then ask!

/Willem

### DIFF
--- a/Processors/TFT_eSPI_STM32.c
+++ b/Processors/TFT_eSPI_STM32.c
@@ -202,7 +202,8 @@ void TFT_eSPI::busDir(uint32_t mask, uint8_t mode)
 #endif
 }
 
-
+// We dont need this. CubeMX will create code for setting the gpiomode.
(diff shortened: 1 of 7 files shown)


wvdakker:
SPI is also integrated and should work.  
The size of the program is from 64k to 48k (based on a STMF746 with 9488 screen).
The touch is the only thing which is not working.

If there are any issues please contact me. Also if some Arduino functions are missing in the
TFT_Arduino library.

PR #3176, opened 2024-02-18
A couple of optimizations/cleanups in TFT_eSPI.cpp

B3rn475:
A couple of optimizations.

- Avoid some checks if previous ones guarantee that they will be false.
- Merge consecutive checks which have the same effect.

### DIFF
--- a/TFT_eSPI.cpp
+++ b/TFT_eSPI.cpp
@@ -2485,8 +2485,7 @@ void TFT_eSPI::fillCircleHelper(int32_t x0, int32_t y0, int32_t r, uint8_t corne
 ***************************************************************************************/
 void TFT_eSPI::drawEllipse(int16_t x0, int16_t y0, int32_t rx, int32_t ry, uint16_t color)
 {
-  if (rx<2) return;
-  if (ry<2) return;
+  if (rx<2 || ry<2) return;
   int32_t x, y;
   int32_t rx2 = rx * rx;
   int32_t ry2 = ry * ry;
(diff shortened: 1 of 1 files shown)


B3rn475:
Note that a lot of this code could be simplified by using `std::min` and `std::max`.

`std::clamp` could also be used in various places, but that would be limited to C++17 and above. On the other hand a local `clamp` helper could be introduced too.

Bodmer:
I suspect these changes have no effect because the compiler will pre-calculate varaibles for inside a loop and minimise comparisons. This means that trivial re-jigging of code is of no benefit.

What code/varaible size and performance difference did you get?

B3rn475:
I've reverted a couple of them, but for the ones that remain:

- I see a reduction of the binary (a few bytes),
- I see differences in the generated asm (jumps are rearrenged to avoid running the checks in the `else if`s),
- I think that some actually improve readability.

B3rn475:
> I suspect these changes have no effect because the compiler will pre-calculate varaibles for inside a loop and minimise comparisons. This means that trivial re-jigging of code is of no benefit.
> 
> What code/varaible size and performance difference did you get?

Is the above comment enough? or should I provide additional data?

Bodmer:
I tried the changes on an RP2040 and it made no difference to binary so I would rather leave the code as is. I expect there are areas where optimisation will give benefits.

B3rn475:
Which example have you tested it against?

The improvement is visible only in examples that use `setTextFont`.

PR #3180, opened 2024-02-19
ST7789: Add mirrored display modes to support 2" rounded rectangle TFT LCD display 240*296 pixels

FedericoBusero:
This pull request adds support for a new 2" TFT LCD display with a rounded rectangle form factor (240*296 pixels) and ST7789 command set.

The display can be used using following settings:

```
#define ST7789_DRIVER
#define TFT_HEIGHT 296
#define CGRAM_OFFSET
#define TFT_RGB_ORDER TFT_BGR
#define TFT_INVERSION_OFF
#define SPI_FREQUENCY  80000000
```
However, the image is mirrored and in some rotations, there is a 24-pixel offset. This pull request addresses these issues by adding support for four additional rotations (4-7) to the ST7789 driver to enable mirroring. In case CGRAM_OFFSET is defined and the height equals 296 pixels, the offset is set when the display is in a mirrored rotation mode.

### DIFF
--- a/TFT_Drivers/ST7789_Rotation.h
+++ b/TFT_Drivers/ST7789_Rotation.h
@@ -1,7 +1,7 @@
   // This is the command sequence that rotates the ST7789 driver coordinate frame
 
   writecommand(TFT_MADCTL);
-  rotation = m % 4;
+  rotation = m % 8;
   switch (rotation) {
     case 0: // Portrait
 #ifdef CGRAM_OFFSET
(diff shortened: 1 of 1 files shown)


cheops (approved):
I tested this change with a screen like [this](https://www.taobao.com/list/item/678831948508.htm)
works ok for me

cheops (approved):
I tested this change with a screen like [this](https://www.taobao.com/list/item/678831948508.htm)
works ok for me

BahuMan (approved):
perfectly backwards compatible

PR #3246, opened 2024-03-22
Add support for Lilygo T-Deck.

hreikin:
Adds support for the [Lilygo T-Deck](https://www.lilygo.cc/products/t-deck) device.


mmuecke:
maybe a help for  #2647

richonguzman:
I tested on another project and confirm that:

1) all pins definition are ok

2)I did not use the following 2 definitions and also all works fine 
 #define TFT_INVERSION_ON
 #define TDECK_PERI_GPIO 10
 
3) I used 
#define SPI_READ_FREQUENCY 20000000 with 16000000 as a value and also works fine

PR #3378, opened 2024-06-22
Ellipse coordinates in int32_t

Marcussacapuces91:
Change x & y coordinates (and color) from int16_t to int32_t as for circle


gbiervliet:
I am no pro so honest question. Why can't you just use the plain "int"?

Marcussacapuces91:
> I am no pro so honest question. Why can't you just use the plain "int"?

Just to be homogeous. "int32_t" everywhere. But you're right, why not "int" or "unsigned" after all coordinates are unsigned int

PR #3393, opened 2024-07-05
Remove HMI gram offset

lewisxhe:
T-HMI needs to remove the coordinate offset to display properly 
https://github.com/Xinyuan-LilyGO/T-HMI/issues/7


Bodmer:
Thanks

Issue #3384, opened 2024-06-30
ESP32 C3 SuperMini ST7735 1.8" screen. All examples fail. Serial Monitor Shows an Error


tgirard:
When I try to use the ESP32 C3 Super Mini with the St7735 Driver, I am never able to get anything on the screen and there seems to be an error that comes up on Example sketches that use the Serial Monitor. My Setup (Pins and Driver) Work fine with the Adafuit driver 
( I really REALLY Don't want to use the Adafruit setup)

I've included:
User_Setup.h (I only changed the driver and pins)
Output from Upload.txt  (This was from 'Output' from the IDE when uploading to the processor)
SerialMonitor Error Information.txt (This is what the Serial Monitor was continuously repeating on Examples that used it)

Let me know if there's anything else I can capture for you :)

IDE: Arduino 2.3.2 Windows 11 
TFT_ESPI: 2.5.43 (I did a clean Install 6/30/2024)
Boards: ESP32 by Espresif 3.0.2
Processor: ESP32 C3 Super Mini From AliExpres
https://www.aliexpress.us/item/3256806413352402.html?spm=a2g0o.order_list.order_list_main.55.5d651802WPL50h&gatewayAdapt=glo2usa
TFT driver: ST7735
Interface type: SPI

I'm more than happy to give you one of these boards if you want one. Just shoot me an address. I'm in the US.

[TFT_ESPI_ERROR_DOCS.zip](https://github.com/user-attachments/files/16045685/TFT_ESPI_ERROR_DOCS.zip)

IAmOrion:
+1 - came here with same problem! 

Additional info...

Serial output:

```
ESP-ROM:esp32c3-api1-20210207
Build:Feb  7 2021
rst:0x3 (RTC_SW_SYS_RST),boot:0xd (SPI_FAST_FLASH_BOOT)
Saved PC:0x40048b82
SPIWP:0xee
mode:DIO, clock div:1
load:0x3fcd5820,len:0x458
load:0x403cc710,len:0x814
load:0x403ce710,len:0x2880
entry 0x403cc710
Guru Meditation Error: Core  0 panic'ed (Store access fault). Exception was unhandled.

Core  0 register dump:
MEPC    : 0x420060a4  RA      : 0x42006094  SP      : 0x3fca2d50  GP      : 0x3fc8d400  
TP      : 0x3fc8e8dc  T0      : 0x00000000  T1      : 0xffc3ffff  T2      : 0x00000000  
[...]

PR #3488, opened 2024-09-21
Fix Boot Loop on ESP32S3 Due to Mis-defined SPI Port

ofthedove:
People have been having trouble using SPI on ESP32 S3. Here are a couple of related discussions:
https://github.com/Bodmer/TFT_eSPI/discussions/3283
https://github.com/espressif/arduino-esp32/issues/10254

The ESP IDF code indicates that only SPI 2 and 3 should be used. See: https://github.com/espressif/esp-idf/blob/9b3eda09741b5c7dae8f22fe60d193dcc3a7ec44/components/soc/esp32s3/include/soc/soc.h#L37

When `SPI_PORT` is defined to `FSPI`, the ESP Arduino framework defines it to `0`. https://github.com/espressif/arduino-esp32/blob/7018cd114d00249674567846c9d67fbb3a1240a3/cores/esp32/esp32-hal-spi.h#L32

Using `0` with `REG_SPI_BASE` in ESP IDF soc.h returns NULL, which causes all subsequent SPI actions to hard-fault.

The simplest solution is to not define `SPI_PORT` to `FSPI`, but rather to GSPI2 or GPSI3 by default, as these are supported by the ESP IDF.

### DIFF
--- a/Processors/TFT_eSPI_ESP32_S3.h
+++ b/Processors/TFT_eSPI_ESP32_S3.h
@@ -80,7 +80,7 @@ SPI3_HOST = 2
   #elif CONFIG_IDF_TARGET_ESP32S2
     #define SPI_PORT 2 //FSPI(ESP32 S2)
   #elif CONFIG_IDF_TARGET_ESP32S3
-    #define SPI_PORT FSPI
+    #define SPI_PORT 3
   #endif
 #endif
 


ofthedove:
@bluecraank Can you share more details? What version of the Arduino ESP32 framework are you on? What error are you getting? Are you confidant that the change is in effect in your application? Are the error messages different after applying the change?

bluecraank:
I have esp32 espressif version 3.0.7. 

The only working fix was in User_Setup.h
 #define USE_HSPI_PORT // HSPI pins 10-13

Versions above 2.0.14 does not work with TFT_eSPI

See my issue https://github.com/espressif/arduino-esp32/issues/10564

Here i found the only working solution right now
https://github.com/Bodmer/TFT_eSPI/issues/3384#issuecomment-2404621623

ofthedove:
Update: I've been running Espressif 32 platform v6.8.1 and framework 2.0.17 (`framework-arduinoespressif32 @ 3.20017.0 (2.0.17)`)

On this version, `USE_HSPI_PORT` and `USE_FSPI_PORT` both work fine.

I just updated to Espressif 32 platform v6.9.0, but this did not change the framework version. I attempted to manually update the framework to version 3.0.7, but it appears this version is incompatible with the v6.9.0 framework. Here's the relevant portion of my platformio.ini:

```
platform = https://github.com/platformio/platform-espressif32.git#v6.9.0
platform_packages =
    framework-arduinoespressif32 @ https://github.com/espressif/arduino-esp32#3.0.7
```

This post has some info on the sort of errors I received when attempting this combination of versions: https://community.platformio.org/t/pio-stopped-working-and-cannot-complie/36236/2

Unfortunately, without being able to use 3.0.7, there's not much more I can do. Perhaps if you could share more info about your platformio.ini config I can investigate further.

casascius:
I spent a bunch of time dealing with this issue today... This change fixed my problem on ESP32 Arduino core version 3 but breaks it on core version 2.  Adding conditional preprocessor logic to only change SPI_PORT to 2 only if ESP_ARDUINO_VERSION_MAJOR >= 3 ended up fixing it for me for now and retained compatibility with version 2 and 3 of ESP32 Arduino core, but this feels like a hack.

PR #3600, opened 2024-12-30
Add Mandelbrot example to DMA folder. It takes code from the 3D spinn…

IanBUK:
…ing cube example, using sprites, and DMA if possible.

The standard Mandelbrot example rendering the set at coordinates (-2.5,-1.25 -> 1.58,1.2) takes just under 21 seconds on my SparkFun Thing and a 800x600 SSD1963 display. Converting it to use sprites renders the same set in 15.5 seconds, the hardware used doesn't allow the full DMA option. This suggests that using sprites is around 25% faster than plotting pixels one-by-one.


IanBUK:
This 'new' example adds the sprite/DMA concept, from the spinning cube example, to the Mandelbrot set example. On my hardware, a Sparkfun Thing and 800x600 SSD1963 display, suggests a 25% speed improvement over the approach of plotting pixels to the screen one-by-one.

PR #3776, opened 2025-07-22
Update GC9A01_Init.h, clear memory before turning the screen on

chriskiefer:
the prevents noise showing on the screen on boot


h4yn0nnym0u5e:
I'd suggest you add some guards around this so it's optional, e.g. ```#if defined(GC9A01_CLEAR_ON_INIT)```. The user can then define the guard in their setup if they want it. I probably wouldn't, it adds even more time to an already-slow startup...

PR #3792, opened 2025-08-24
Fix issue with ESP32 C3 SPI not working on newer Arduino IDE

themrleon:
After struggling to get the ESP32 C3 Super Mini working with a ST7789 TFT SPI I found this thread that solved, it seems the lib needs an update to keep working with latest Arduino versions, solution provided by "Moriyama":

For details and context please check:
https://esp32.com/viewtopic.php?t=37617&start=20#p142493

### DIFF
--- a/Processors/TFT_eSPI_ESP32_C3.h
+++ b/Processors/TFT_eSPI_ESP32_C3.h
@@ -68,7 +68,11 @@ SPI3_HOST = 2
 */
 
 // ESP32 specific SPI port selection - only SPI2_HOST available on C3
+#if ESP_ARDUINO_VERSION_MAJOR < 3
 #define SPI_PORT SPI2_HOST
+#else
+#define SPI_PORT 2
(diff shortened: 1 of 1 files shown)


reparator:
Thanks Leonardo,
finding this fix helped me get my combination ESP32-C3-Super Mini and IPS 135x240 ST7789 TFT working!

reparator (changes requested):
This fix works!

Bodmer:
Good call. I am making other changes so will wrap this in.

Documentation clarifications have revealed that with the C3 only SPI2 port is available to users, both SPI0 & SPI1 are dedicated to internal flash memory hence the issue.

I will leave this here until the changes are wrapped in.

noliran:
+1 on this fix :)
though I'm not sure why it's conditioned on `ESP_ARDUINO_VERSION_MAJOR`

PR #3814, opened 2025-10-14
Update ILI9488_Rotation.h

SebastianLang-GER:
Add support for mirrored rotation (4-7) in ILI9488 driver.

### DIFF
--- a/TFT_Drivers/ILI9488_Rotation.h
+++ b/TFT_Drivers/ILI9488_Rotation.h
@@ -1,7 +1,7 @@
   // This is the command sequence that rotates the ILI9488 driver coordinate frame
 
   writecommand(TFT_MADCTL);
-  rotation = m % 4;
+  rotation = m % 8;
   switch (rotation) {
    case 0: // Portrait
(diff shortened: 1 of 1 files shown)


Bodmer:
I see no harm in adding this but please explain the need for the change. Why did you need to make the change, i.e. what problem does it solve?

SebastianLang-GER:
> I see no harm in adding this but please explain the need for the change. Why did you need to make the change, i.e. what problem does it solve?

I had a small series of about 50 displays with ILI9488 chip that should have been all the same, but 6 of them showed mirrored screens.
I tried a lot with the available options and compared the code to other drivers in your wonderful library.
By adding the **mirrored rotation options (4-7)** in the same way it is done in several other drivers, I could make everything appear correctly for all my displays.

So the concept of the code is not new (see other drivers) and It was fully tested by me.
With this pull request I hope others with the same problem find an easy solution.

Best regards

Bodmer (approved):
All good.

Bodmer:
Thanks.

PR #3818, opened 2025-10-16
Added config for the LilyGo T-Display S3

savigny030:
Just used the Setup template and added all pins for the LilyGo T-Display S3 (ESP32-S3) pinout.


Bodmer:
Can you tidy the setup file to the bare essentials:

1 Remove irrelevant comments 
2 remove confusing SPI pin defines. It is a parallel bus interface?

I think that at the moment it will not work(!) the following line is missing? :
#define TFT_PARALLEL_8_BIT

Check for error messages and confirm correct operation with a the target board.

Thanks.

PR #3827, opened 2025-10-28
Add PC simulator support

shuai132:
add simulator for pc platform, using SDL2 render.

image


Bodmer:
This is a good idea... but it will need a user manual/tutorial on how to make use of this capability. Since I have no experience of using the SDL2 tools I would not be able to support users to sort out problems.

So thanks for the pull request but I think this might cause me problems when support or bugs are discovered.

PR #3832, opened 2025-11-05
dma queueing 

piotrcurious:


piotrcurious:
nah, no buffer persistence

PR #3876, opened 2026-02-09
Update TFT pin definitions for new configuration

takaabiezer-stack:


Bodmer:
The idea is to add you own setups in User_Setups and link to them in User_Setup_Select.h

PR #3877, opened 2026-02-09
Create setup1.h for TFT LCD configuration

takaabiezer-stack:
Add configuration header for 3.5 inch TFT LCD.


Bodmer:
Given the infinite variations I will only be adding setups I have personally tested.

Keep a copy handy if you ever update TFT_eSPI as it will get overwriiten.

PR #3886, opened 2026-02-13
Add STM32U5 support: register-level SPI and GPDMA DMA

welcometotheroot:
- Add U5 SPI register mappings (TXDR/TXP/TXC vs DR/TXE/BSY)
- Add SET_BUS_WRITE_MODE for U5 unlimited streaming (TSIZE=0)
- Add GPDMA-based DMA for pushPixelsDMA/pushImageDMA on U5
- Add both GPDMA and SPI IRQ handlers (U5 two-stage completion)
- Guard SPI_BUSY_CHECK with SPE to handle post-DMA disabled state
- Add DMA stub fallback for STM32 variants without DMA support


welcometotheroot:
For larger screens this implementation isn't efficient. We should be using the linked-list DMA support. Will put this in draft until that's done

PR #3891, opened 2026-02-18
Update User_Setup.h

stansajak:


HackerGoose3000:
Just wanted to add that as of May 2026 the ST7789 240x240  is also confirmed working on the ESP32 Dev Module by myself and another user. I say this because there is a line in there saying...
// For ESP32 Dev board (only tested with ILI9341 display)

I posted a full breakdown of gotchas for ST7789 + SD card  on ESP32 in the Discussions section if anyone needs it.

PR #3892, opened 2026-02-18
Update User_Setup.h

stansajak:


stansajak:
OK

stansajak:
Zmiana numeró pin

stansajak:
OK

tochinet:
This is obviously not the way to go, but a better approach would be to define the pins in the .ino file (or at least test with #ifdef before defining in the library. Look at how it is done in https://github.com/mobizt/ReadyMail/blob/main/examples/Sending/ESP32Camera/ESP32Camera.ino for an example. I dropped tft_espi because of this serious caveat : the library is always broken when you change board

PR #3899, opened 2026-03-05
Add Setup73 for ESP32-S3 with ST7789V 240x320

uhojin:
## Summary
- Add `Setup73_ESP32_S3_ST7789V.h` user setup for ESP32-S3 with ST7789V 240x320 SPI display
- Add corresponding entry in `User_Setup_Select.h`

## Details
- Tested with Freenove ESP32-S3-WROOM (N8R8) and Waveshare 2" SPI LCD
- Uses FSPI default pins (GPIO 10, 11, 12) for optimal SPI performance
- DC (GPIO 14) and RST (GPIO 21) chosen to avoid conflicts with onboard camera pins on Freenove boards

## Screenshot
demo


uhojin:
Hello @Bodmer! Thanks for the amazing library.

I couldn't find any guidelines/patterns on assigning ID for the setup so I just used 70s range + available number.
Please let me know if there's a rule for assigning Setup ID, and I will update it accordingly.

PR #3930, opened 2026-06-20
Update TOUCH_CS pin definition in User_Setup.h

Sujitbadhiye21:
change #define touch_CS 21 into   _ CS 5


Sujitbadhiye21:
change

